PR #30, opened 2023-05-08
Fix really to support publish intent

TimEvens:


cmd/really/reallyTest.cpp
```diff
@@ -139,7 +140,11 @@ main(int argc, char* argv[])
   quicr::QuicRClient client(relay, tcfg, logger);
 
   if (data.size() > 0) {
-    // do publish
+    auto nspace = quicr::Namespace(name, 96);
+    client.publishIntent(pd, nspace, {}, {}, {});
```
TimEvens, line 144: Need to issue publish intent to allow message forwarding.

paulej (approved):

GhostofCookie (approved):

GhostofCookie:
Looks like Ubuntu is mad:
```log
/home/runner/work/libquicr/libquicr/cmd/really/really.cpp:221:94: error: designator order for field ‘Subscriptions::Remote::subscribe_id’ does not match declaration order in ‘Subscriptions::Remote’

TimEvens:
> 

Classic order issue. :(   Updated.  Problem is that IDE doesn't catch this all the time and macos allows it.

PR #33, opened 2023-05-16
Restructured code to facilitate adding session layers

paulej:


include/quicr/quicr_client_delegate.h
```diff
@@ -0,0 +1,143 @@
...
+
+#include <cstdint>
+#include "quicr/quicr_name.h"
+#include "quicr/quicr_namespace.h"
+#include "quicr/quicr_common.h"
+
+namespace quicr {
+
+/*
+ *  Subscriber delegate callback methods
```
suhasHere, line 26: s/Subscriber delegate callback methods/ QUICR Subscriber delegate callback methods/

include/quicr/quicr_client_session.h
```diff
@@ -0,0 +1,164 @@
+/*
+ *  quicr_client_session.h
+ *
+ *  Copyright (C) 2023
+ *  Cisco Systems, Inc.
+ *  All Rights Reserved
+ *
+ *  Description:
+ *      This is an interface specification for the session layer sitting at the
+ *      next level of the client library.
```
suhasHere, line 10: may be clarify what is the next level here by drawing an ascii layer cake

include/quicr/quicr_server_session.h
```diff
@@ -0,0 +1,106 @@
...
+ *  quicr_server_session.h
+ *
+ *  Copyright (C) 2023
+ *  Cisco Systems, Inc.
+ *  All Rights Reserved
+ *
+ *  Description:
+ *      This is an interface specification for the session layer sitting at the
+ *      next level of the server library.
+ *
```
suhasHere, line 11: same comment as ealrier, can we clarify what does next level means

src/quicr_client_raw_session.h
```diff
@@ -0,0 +1,251 @@
+/*
+ *  quicr_client_raw_session.h
+ *
+ *  Copyright (C) 2023
+ *  Cisco Systems, Inc.
+ *  All Rights Reserved
+ *
+ *  Description:
+ *      This file defines a session layer between the client APIs and the
+ *      transport that uses raw data packets (either UDP or QUIC).
```
suhasHere, line 10: please add a note to clarify what does raw data packets mean for UDP or QUIC in the context of session

src/quicr_client_raw_session.h
```diff
@@ -0,0 +1,251 @@
...
+#include "quicr/quicr_client_delegate.h"
+#include "quicr/quicr_client.h"
+
+namespace quicr {
+
+/**
+ *   Client Raw Session Interface
+ */
+class QuicRClientRawSession : public QuicRClientSession
+{
```
suhasHere, line 37: may be rename it to QuicRClientNativeSession ?

src/quicr_client_raw_session.h
```diff
@@ -0,0 +1,251 @@
...
+                                  bool is_last_fragment,
+                                  bytes&& data) override;
+
+  void handle(messages::MessageBuffer&& msg);
+  void removeSubscribeState(bool all, const quicr::Namespace& quicr_namespace,
+                            const SubscribeResult::SubscribeStatus& reason);
+
+  std::shared_ptr<qtransport::ITransport> transport;
+  qtransport::LogHandler& log_handler;
+
```
suhasHere, line 190: any reason these are public ?

src/quicr_client_raw_session.h
```diff
@@ -0,0 +1,251 @@
...
+protected:
+  std::mutex mutex;
```
suhasHere, line 192: mutex has a variable name is confusing, can we please rename it to idetify its purpose ?

src/quicr_server_raw_session.cpp
```diff
@@ -0,0 +1,509 @@
...
+ *  quicr_server_raw_session.h
+ *
+ *  Copyright (C) 2023
+ *  Cisco Systems, Inc.
+ *  All Rights Reserved
+ *
+ *  Description:
+ *      This file implements a session layer between the client APIs and the
+ *      transport that uses raw data packets (either UDP or QUIC).
+ *
```
suhasHere, line 11: please clarify what does raw data packets mean?

src/quicr_server_raw_session.h
```diff
@@ -0,0 +1,232 @@
...
+
+
+
+  private:
+    QuicRServerRawSession& server;
+  };
+
+public:
+  std::mutex mutex;
+
```
suhasHere, line 158: please rename mutex

suhasHere (changes requested):
@paulej  this looks like a fine start. I have few editorial nits.

include/quicr/quicr_client_delegate.h
paulej: Will change.

include/quicr/quicr_client_session.h
paulej: I'll add a topology comment.

include/quicr/quicr_server_session.h
paulej: Ditto.

src/quicr_client_raw_session.h
paulej:
```cpp
 *      transport that uses raw data packets, namely UDP or QUIC.
```

I'll change it to that.  Raw just means UDP or QUIC.  The other one coming is quicr_client_h3_session.h.  It will make more sense with that present.

src/quicr_client_raw_session.h
paulej: What is Native?  These are raw packets on the wire, as opposed to being wrapped in HTTP/3 or something.

src/quicr_server_raw_session.cpp
paulej: Yeah, whoever wrote the code did so in such a way that they needed access to these.  It's the way the transport callback code is written.

src/quicr_client_raw_session.h
suhasHere: yes, i do understand the context. My comment for someone external who look at the header, it is lacking context on being raw .. a note on the context will help

src/quicr_server_raw_session.h
paulej: Done.

src/quicr_client.cpp
```diff
@@ -1,665 +1,113 @@
...
-  messages::MessageBuffer msg{ sizeof(messages::PublishIntent) +
-                               payload.size() };
-  msg << intent;
-
-  auto error =
-    transport->enqueue(transport_context_id, transport_stream_id, msg.get());
-
-  return error == qtransport::TransportError::None;
+  return client_session->publishIntent(
+    pub_delegate, quicr_namespace, origin_url, auth_token, std::move(payload));
```
GhostofCookie, line 45: I think std::move might be redundant here, but I don't disapprove of being explicit.

suhasHere (approved):

src/quicr_client.cpp
```diff
@@ -1,665 +1,113 @@
...
 }
 
-QuicRClient::~QuicRClient()
+QuicRClient::QuicRClient(std::shared_ptr<qtransport::ITransport> transport_in)
 {
-  removeSubscribeState(true, {},
-                      SubscribeResult::SubscribeStatus::ConnectionClosed);
-  transport.reset(); // wait for transport close
+  client_session =
+    std::make_unique<QuicRClientRawSession>(transport_in);
```
GhostofCookie, line 34: So, it's worth noting that QuicrClient has become mostly a wrapper of QuicrSession. THis isn't necessarily a bad thing, but since these 2 constructors are the only real difference, I'd like to make a pitch:

Why not do away with QuicrClient/QuicrServer, and instead turn these into standard library style make methods:
```cpp
std::shared_ptr<quicr::ClientSession> make_client_session(...);
std::shared_ptr<quicr::ServerSession> make_server_session(...);
```

src/quicr_client_raw_session.h
GhostofCookie: I think Native sounds misleading, since it is usually used to refer to something in relation to an environment more than a mode.

src/quicr_client_raw_session.h
GhostofCookie: +1

These members should be private, or at least protected.

src/quicr_client_raw_session.h
```diff
@@ -0,0 +1,251 @@
...
+
+protected:
+  std::mutex session_mutex;
+
+  bool notify_pub_fragment(const messages::PublishDatagram& datagram,
+                           const std::map<int, bytes>& frag_map);
+  void handle_pub_fragment(messages::PublishDatagram&& datagram);
+
+  qtransport::LogHandler def_log_handler;
+
```
GhostofCookie, line 199: The members are a little buried in here, could these be moved further down past the function and type declarations?

src/quicr_client_raw_session.h
paulej: Yeah, whoever wrote the code did so in such a way that they needed access to these. It's the way the transport callback code is written.

src/quicr_client_raw_session.h
paulej: Agreed. This is how the code was.  I didn't try to improve it. I just restructured so I can insert a different session.  This (among other things) needs to be revisited.

src/quicr_client.cpp
paulej: Compiler doesn't agree :)

PR #39, opened 2023-06-15
Add streaming modes (Experimental)

GhostofCookie:


src/quicr_client_raw_session.cpp
```diff
@@ -174,7 +205,9 @@ QuicRClientRawSession::make_transport(RelayInfo& relay_info,
     throw std::runtime_error(log_msg.str());
   }
 
-  transport_stream_id = transport->createStream(transport_context_id, false);
+  _stream_mode = stream_mode;
+  _control_stream_id =
+    transport->createStream(transport_context_id, is_reliable());
```
GhostofCookie, line 210: I think this needs one previous call to createStream with unreliable, but may not need to save off the datagram stream.

src/quicr_server_raw_session.cpp
```diff
@@ -274,38 +324,54 @@ QuicRServerRawSession::handle_unsubscribe(
...
+    context = states[context_id];
   }
 
-  delegate.onPublisherObject(context.transport_context_id,
-                             context.transport_stream_id,
-                             false,
-                             std::move(datagram));
+  // Check if this is the datagram stream, and react accordingly if not.
+  bool new_stream = false;
+  if (stream_id >= 4) {
```
TimEvens, line 353: We have been leveraging that stream ID must start at 1 for a sequence number. This resulted in the `stream_id` always being `>= 4`.   It does simplify code a bit.... but RFC9000 only states that 0  - `2**62-1` are valid, leaving 0, 1, 2, 3 as valid stream IDs.   We have been using stream ID 0 and 1 to indicate datagram.  Picoquic has different callbacks, so it's possible we add to the context if the message is datagram or stream based. This would allow us to support stream_ids being 0, 1, 2, 3.

src/quicr_server_raw_session.cpp
```diff
@@ -317,21 +383,19 @@ QuicRServerRawSession::handle_publish_intent(
   messages::PublishIntent intent;
   msg >> intent;
 
-  if (!publish_namespaces.count(intent.quicr_namespace)) {
-    PublishIntentContext context;
-    context.state = PublishIntentContext::State::Pending;
+  if (!publish_intent_state.count(context_id)) {
+    PublishContext context;
+    context.state = PublishContext::State::Pending;
```
TimEvens, line 388: Not related to this PR, but we could move this to LAPS.

TimEvens (changes requested):
Considering the change might introduce other unknown issues, we might want to hold off on this till after next week.  

Overall the changes look good.  There will be some merge conflicts though due to
priority queue updates that are needed sooner.  :( 

`QuicRServer` should also have `_control_stream_id`.  

Maybe break out `quicr_namespace.h` to a new PR?.?

src/helpers.h
```diff
@@ -0,0 +1,31 @@
+#include "quicr/quicr_common.h"
+
+#include <transport/transport.h>
+
+namespace {
+constexpr quicr::RelayInfo::Protocol
```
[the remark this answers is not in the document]
GhostofCookie: This header is not exposed to clients.

src/quicr_server_raw_session.cpp
```diff
@@ -15,16 +15,55 @@
...
+ * Based on the name changing, checks whether the difference of the last name
+ * and the new name is a change in the group id bits, or the object id bits, and
+ * returns accordingly. However, if the bits have changed for that stream mode,
+ * but the stream is the same, then we have to consider the possiblities:
+ *  1. If the object is updating, but the stream is not, then the mode could be
+ *     PerGroup
+ *  2. If the group id is updating, but the stream is the same, then the mode
+ *     could be PerPriority
+ *
+ * @param same_stream
```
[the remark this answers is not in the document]
GhostofCookie: Oops!

src/quicr_server_raw_session.cpp
```diff
@@ -15,16 +15,55 @@
...
+ */
+quicr::StreamMode
+detect_stream_mode(bool same_stream,
+                   quicr::Name last_name,
+                   quicr::Name new_name)
+{
+  quicr::StreamMode interim_mode = quicr::StreamMode::Datagram;
+
+  auto diff = new_name - last_name;
+  if (diff < (0x1_name << 48)) {
```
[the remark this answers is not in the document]
GhostofCookie: Yeah, magic numbers right now. Not a fan. Could give them names for the meantime, but should definitely be something configurable.

src/quicr_server_raw_session.cpp
GhostofCookie: Aha! I knew I left something weird in here somewhere. Good catch!

src/quicr_server_raw_session.cpp
```diff
@@ -111,26 +150,31 @@ QuicRServerRawSession::run()
 void
 QuicRServerRawSession::publishIntentResponse(
   const quicr::Namespace& quicr_namespace,
+  const qtransport::TransportContextId& context_id,
   const PublishIntentResult& result)
 {
-  if (!publish_namespaces.count(quicr_namespace))
+  if (!publish_intent_state.count(context_id))
```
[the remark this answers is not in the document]
GhostofCookie: The reason I kept it is because the way I am using publish_state, the publish intent context could be overridden by the next publish object. Though maybe it's a non-issue, since I think I changed it so that the publish namespace isn't added until a successful publishIntentResponse (effectively blocking published named objects going through yet)... so might be safe now. I'll poke around with that again.

src/quicr_client_raw_session.cpp
```diff
@@ -359,42 +388,44 @@ QuicRClientRawSession::unsubscribe(const quicr::Namespace& quicr_namespace,
...
     log_handler.log(qtransport::LogLevel::info, log_msg.str());
     return;
   }
 
   auto& [ns, context] = *found;
 
-  context.group_id = datagram.header.name.bits<uint64_t>(16, 32);
-  context.object_id = datagram.header.name.bits<uint64_t>(0, 16);
+  context.group_id = name.bits<uint64_t>(16, 32);
+  context.object_id = name.bits<uint64_t>(0, 16);
```
GhostofCookie, line 422: Hard coded values, will make dynamic later.

cmd/really/really.cpp
```diff
@@ -141,16 +141,16 @@ class ReallyServer : public quicr::ServerDelegate
   ~ReallyServer() { server.reset(); };
 
   virtual void onPublishIntent(const quicr::Namespace& quicr_namespace,
+                               const qtransport::TransportContextId context_id,
                                const std::string& /* origin_url */,
-                               bool /* use_reliable_transport */,
```
suhasHere, line 145: I don't think we should remove this.

GhostofCookie: I'd be inclined to agree, though I have to ask why? The way it is right now the publish intent is sent on the control stream, which is already set to reliable/unreliable.

src/quicr_server_raw_session.cpp
GhostofCookie: This is still in the code right now, guidance on a better way for this is appreciated.

GhostofCookie:
Closing because it is too far behind and not really reflective of the direction we've moved.

PR #41, opened 2023-06-17
Pass priority and TTL to transport

TimEvens:
This appears to be working okay with WxQ/new-qmedia, but the TTL will need some adjusting in conf server/manifest. Right now the TTL is set to 50ms, which should be fine, but apparently it's not fine.  It needs to be more like 100ms. The problem appears to be with picoquic taking too long a times that exceeds 50ms.  That's a bit crazy, so we need to dig into that. Regardless, that's unrelated to the pq/time-queue changes here.

### DIFF
--- a/cmd/really/really.cpp
+++ b/cmd/really/really.cpp
@@ -149,6 +149,7 @@ public:
                                quicr::bytes&& /* e2e_token */)
   {
     // TODO: Authenticate token
+    logger.log(qtransport::LogLevel::info, "Publish intent namespace: " + quicr_namespace.to_hex());
     quicr::PublishIntentResult result{ quicr::messages::Response::Ok, {}, {} };
     server->publishIntentResponse(quicr_namespace, result);
   };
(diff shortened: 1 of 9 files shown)


cmd/really/really.cpp
```diff
@@ -176,7 +177,7 @@ class ReallyServer : public quicr::ServerDelegate
         continue;
       }
 
-      server->sendNamedObject(dest.subscribe_id, false, datagram);
+      server->sendNamedObject(dest.subscribe_id, false, 1, 200, datagram);
```
TimEvens, line 180: The server side does not forward priority and ttl yet because we need to change the message header to support that. I plan to do that next week during the offsite in a different commit/PR.

suhasHere (approved):

suhasHere:
> . Right now the TTL is set to 50ms, which should be fine, but apparently it's not fine. It needs to be more like 100ms.

both seems to be really small TTL values. I would think it should be 500ms or more .. may be 1 sec

GhostofCookie (approved):

PR #42, opened 2023-06-19
Updating name to do some name arithmetic and added namesapce map.

GhostofCookie:
Changes from #39 , only includes the name stuff, since this can safely go to main.


TimEvens (approved):
LGTM, but I'm not sure of the use-case for `name1 [+-] name2` is...

PR #45, opened 2023-06-22
fetch message v1

suhasHere:
This is a v1 version of GET API  as a starting point. Once we experiment with the flows, we may want to revisit the right format and encoding of the same.

I will work on adding support for GET in the client/server API next

### DIFF
--- a/include/quicr/encode.h
+++ b/include/quicr/encode.h
@@ -192,33 +192,19 @@ operator>>(MessageBuffer& msg, Namespace& ns);
 
 
 /*===========================================================================*/
-// GET Message Types
+// Fetch Message Types
 /*===========================================================================*/
 
-struct Get
+struct Fetch
 {
   uint64_t transaction_id;
-  quicr::Namespace  resource; // resource to retrieve
+  quicr::Name name; // resource to retrieve
   // TODO - Add authz
 };
 
 MessageBuffer&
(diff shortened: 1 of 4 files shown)


include/quicr/encode.h
```diff
@@ -190,4 +190,35 @@ operator<<(MessageBuffer& msg, const Namespace& ns);
...
+MessageBuffer&
+operator>>(MessageBuffer& buffer, Get& msg);
+
+struct GetResponse
+{
+  quicr::Namespace resource;
+  SubscribeResult::SubscribeStatus response;
+  uint64_t transaction_id; // matches the get request
+  uintVar_t num_objects;
+  std::vector<PublishDatagram> objects;
```
GhostofCookie, line 216: In reference to your TODO later, maybe this should be a vector of PublishStream, and you can add a Header field to this struct?

GhostofCookie (approved):

include/quicr/quicr_common.h
```diff
@@ -42,6 +42,8 @@ enum class MessageType : uint8_t
   PublishIntent,
   PublishIntentResponse,
   PublishIntentEnd,
+  Get,
+  GetResponse,
```
GhostofCookie: Should these be `Fetch` now since the name of the message has changed?

PR #50, opened 2023-06-30
Improving speed of MessageBuffer writes.

GhostofCookie:
Okay, this PR started off as me messing around on my time off. It ended up becoming a pretty large thing, though I think it's made MessageBuffer stronger.

The big parts here are that:
- Name and consequently Namespace are cheaper to write
- Message types and uintVar_t are in their own respective headers
- encode.h has been turned into what's its name implies: a header with encode/decode methods
- MessageBuffer has become more concrete, and hopefully slightly faster
- Span is being used now when passing vectors to write to the buffer, and when access an immutable view/slice of the buffer (to reduce copy)

For sure Name is noticeably cheaper, going from 34-40ns on average to write to the buffer before to 5-8ns on average (all on my machine, GH Actions tells a similar story though)

### DIFF
--- a/include/quicr/message_buffer.h
+++ b/include/quicr/message_buffer.h
@@ -62,6 +62,76 @@ private:
 
 namespace quicr::messages {
 
+namespace {
+
+template<typename T>
+constexpr T
+swap_bytes(T value)
+{
+  if constexpr (std::endian::native == std::endian::big)
+    return value;
+
+  uint8_t* value_ptr = reinterpret_cast<uint8_t*>(&value);
+  for (size_t i = 0; i < sizeof(value) / 2; ++i) {
+    std::swap(value_ptr[i], value_ptr[sizeof(value) - 1 - i]);
+  }
+  return value;
+}
+
+template<>
+constexpr uint16_t
+swap_bytes(uint16_t value)
+{
+  if constexpr (std::endian::native == std::endian::big)
+    return value;
+
+  return ((value >> 8) & 0x00ff) | ((value << 8) & 0xff00);
+}
+
+// clang-format off
+template<>
+constexpr uint32_t
+swap_bytes(uint32_t value)
+{
+  if constexpr (std::endian::native == std::endian::big)
+    return value;
+
+  return ((value >> 24) & 0x000000ff) |
+         ((value >>  8) & 0x0000ff00) |
+         ((value <<  8) & 0x00ff0000) |
+         ((value << 24) & 0xff000000);
+}
+
+template<>
+constexpr uint64_t
+swap_bytes(uint64_t value)
+{
+  if constexpr (std::endian::native == std::endian::big)
+    return value;
+
+  return ((value >> 56) & 0x00000000000000ff) |
+         ((value >> 40) & 0x000000000000ff00) |
+         ((value >> 24) & 0x0000000000ff0000) |
+         ((value >>  8) & 0x00000000ff000000) |
+         ((value <<  8) & 0x000000ff00000000) |
+         ((value << 24) & 0x0000ff0000000000) |
+         ((value << 40) & 0x00ff000000000000) |
+         ((value << 56) & 0xff00000000000000);
+}
+
+template<>
+constexpr quicr::Name
+swap_bytes(quicr::Name value)
+{
+  if constexpr (std::endian::native == std::endian::big)
+    return value;
+
+  constexpr auto ones = ~0x0_name;
+  return ((ones & swap_bytes(uint64_t(value))) << 64) | swap_bytes(uint64_t(value >> 64));
+}
+}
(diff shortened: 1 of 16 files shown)


include/quicr/message_buffer.h
```diff
@@ -106,87 +104,145 @@ class MessageBuffer
...
+  /**
+   * @brief Overload for Namespace integral type to write in NBO.
+   * @param msg The MessageBuffer to write to.
+   * @param value The message to be written.
+   * @returns The MessageBuffer that was written to.
+   */
+  friend MessageBuffer& operator<<(MessageBuffer& msg, quicr::Namespace value)
+  {
+    if constexpr (std::endian::native == std::endian::big)
+      return msg << value.name() << value.length();
```
paulej: These should be written the same regardless of architecture. The actual values might have a different byte order.

include/quicr/message_buffer.h
```diff
@@ -106,87 +104,145 @@ class MessageBuffer
...
-    return value;
+  /**
+   * @brief A fancy operator for pop, reads a byte off the buffer.
+   * @param buffer The MessageBuffer to read from.
+   * @param value The value to read into.
+   * @returns The MessageBuffer that was read from.
+   */
+  friend inline MessageBuffer& operator>>(MessageBuffer& msg, uint8_t& value)
+  {
+    if (msg.empty()) {
```
paulej: While this addresses an empty buffer, since some calls to front() might demand more data than the buffer has, why not just have an exception thrown from front()?  It would reduce how many places one throws an exception.

include/quicr/message_buffer.h
```diff
@@ -106,87 +104,145 @@ class MessageBuffer
...
-}
+    auto val_ptr = reinterpret_cast<uint8_t*>(&value);
+    std::memcpy(val_ptr, msg._buffer.data(), sizeof(T));
 
-constexpr quicr::Name
-swap_bytes(quicr::Name value)
-{
-  if constexpr (std::endian::native == std::endian::big)
-    return value;
+    msg._buffer.erase(msg._buffer.begin(),
```
paulej: I suppose this is one reason to have exceptions in these friend classes, though it would be nice if one could just call `ReadValue()` in this case and have the exception thrown in the MessageBuffer.  This is what Cantina's DataBuffer does.  It lacks these operators, but it addresses this sort of issue.  Also, doing this erase() is truly unfortunate, as it causes memmove() repeatedly.  In Cantina's DataBuffer, we keep references to the current read position, as opposed to moving memory.

include/quicr/message_buffer.h
```diff
@@ -106,87 +104,145 @@ class MessageBuffer
...
+   * @param value The value to read into.
+   * @returns The MessageBuffer that was read from.
+   */
+  friend MessageBuffer& operator>>(MessageBuffer& msg, quicr::Namespace& value)
+  {
+    quicr::Name name;
+    uint8_t length;
+
+    if constexpr (std::endian::native == std::endian::big)
+      msg >> name >> length;
```
paulej: Order of elements should be the same here, too, regardless of architecture

include/quicr/message_buffer.h
GhostofCookie: We could absolutely do that. We did that in time_queue, could implement it here easily as well.

paulej (approved):

src/message_buffer.cpp
```diff
@@ -45,64 +66,62 @@ MessageBuffer::push(std::vector<uint8_t>&& data)
...
 {
-  if (len == 0)
+  if (length == 0)
     return {};
 
-  if (len > _buffer.size())
-    throw OutOfRangeException(
-      "len cannot be longer than the size of the buffer");
+  if (length > size())
+    throw OutOfRangeException(length, size());
```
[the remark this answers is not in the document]
paulej: Withdrawn. I read it wrong.

src/quicr_client_raw_session.cpp
```diff
@@ -123,9 +123,10 @@ class QuicRTransportDelegate : public qtransport::ITransport::TransportDelegate
                                "Dropping malformed message: " +
                                  std::string(e.what()));
         return;
-      } catch (const std::exception& /* ex */) {
+      } catch (const std::exception& e) {
         client.log_handler.log(qtransport::LogLevel::info,
-                               "Dropping malformed message");
+                               "Dropping malformed message: " +
+                                 std::string(e.what()));
```
paulej, line 129: While that might work, why not put `std::string()` around the C-string? Then the `operator+` will work. Not a big deal, if this gets converted properly.

paulej (approved):
I thought much of this was already merged. I can't tell what's new or old.  But, if it's faster, good. :)

PR #51, opened 2023-07-11
Fixing using wrong name bits.

GhostofCookie:


shenning00 (approved):

RichLogan (approved):
Would it make sense (maybe subsequently) to have some `quicr_name::get_object_id()` or `quicr_name::get_group_id()` functions/extensions where we need to make that distinction?

PR #53, opened 2023-07-21
Fix issue related to crash when erasing fragments

paulej:
This may address the issue raised in https://github.com/Quicr/WxQ/issues/152


TimEvens (approved):
LGTM

RichLogan (approved):
I can't seem to reproduce the issue in the wild, but as I understand it this fix makes sense.

PR #56, opened 2023-07-27
Cleaning up client session arch to stop frag crash.

GhostofCookie:
A bit of context here: The crashes around `notify_pub_fragment` (an example shown in https://github.com/Quicr/WxQ/issues/152) seem to be caused by a race with deleting entries from the `sub_delegates` map, and trying to access the whole map in the notify.

A few things here:
- The `subscribe_state` map was being fully emptied by both the `QuicrClientRawSession` and `QuicrClientTransportDelegate` destructors, which doesn't make a whole lot of sense.
- The QuicrClientTransportDelegate was acting as a hollow shell of the session itself (one could argue this was the point of a delegate), but didn't actually do any real work itself, just turned around and called the session to do the work anyways (not very delegate-like).

The changes here:
- QuicrRawClientSession becomes the TransportDelegate itself, so it doesn't abstract away calls to itself to a member owned by it (which consequently also stored a reference to the session, very circular).
- The order of the members in the QuicrRawClientSession class are reordered to ensure a more proper teardown order (removing the need to explicitly call reset on a pointer).
- Changing up connection status handling a little bit, and allowing a stop to be added to cancel connecting.

### DIFF
--- a/src/quicr_client_raw_session.cpp
+++ b/src/quicr_client_raw_session.cpp
@@ -13,6 +13,14 @@
  *      None.
  */
 
+#include "quicr_client_raw_session.h"
+
+#include "quicr/encode.h"
+#include "quicr/message_buffer.h"
+#include "quicr/quicr_client.h"
+#include "quicr/quicr_client_delegate.h"
+#include "quicr/quicr_common.h"
+
 #include <chrono>
 #include <iomanip>
 #include <iostream>
@@ -20,15 +28,9 @@
 #include <string>
 #include <thread>
 
-#include "quicr/encode.h"
-#include "quicr/message_buffer.h"
-#include "quicr/quicr_client.h"
-#include "quicr/quicr_client_delegate.h"
-#include "quicr/quicr_common.h"
-#include "quicr_client_raw_session.h"
-
 namespace quicr {
 
+namespace {
 /*
  * Nested map to reassemble message fragments
  *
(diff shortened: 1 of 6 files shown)


src/quicr_client_raw_session.cpp
```diff
@@ -37,184 +39,199 @@ namespace quicr {
...
+{
   transport_context_id = transport->start();
 
+  log_handler.log(
+    qtransport::LogLevel::info,
+    (std::ostringstream() << "Waiting for client to be ready...").str());
+
   client_status = ClientStatus::CONNECTING;
+  while (client_status == ClientStatus::CONNECTING) {
+    if (stopping)
```
GhostofCookie, line 124: This doesn't quite do the trick yet.

src/quicr_client_raw_session.cpp
```diff
@@ -37,184 +39,240 @@ namespace quicr {
...
-            << int(transport->status());
-    logger.log(qtransport::LogLevel::info, log_msg.str());
+  while (!stopping && client_status == ClientStatus::CONNECTING) {
     std::this_thread::sleep_for(std::chrono::milliseconds(100));
   }
 
-  client_status = ClientStatus::READY;
+  if (stopping) {
+    log_handler.log(qtransport::LogLevel::info,
+                    (std::ostringstream() << "Cancelling connecting session "
```
GhostofCookie, line 128: Though we try hard, picoquic (and even harder, UDP) decide that they don't want to be stopped, and continue on without a care. So fully stopping this when we want to leave a call if we didn't connect is a bit of an issue.

paulej (approved):

src/quicr_client_raw_session.cpp
```diff
@@ -539,36 +578,31 @@ bool
...
 
-    reassembled.insert(
-      reassembled.end(), item.second.begin(), item.second.end());
-    seq_bytes += item.second.size();
-  }
+    reassembled.insert(reassembled.end(),
+                       std::make_move_iterator(data.begin()),
+                       std::make_move_iterator(data.end()));
 
-  for (const auto& entry : sub_delegates) {
```
TimEvens, line 562: Good to remove this as it was old code for the idea that we would have multiple matches.

TimEvens (approved):
LGTM

src/quicr_client_raw_session.cpp
GhostofCookie: Created https://github.com/Quicr/libquicr/issues/57 for this.

PR #70, opened 2023-08-31
Changed libquicr to use the Cantina Logger

paulej:
This uses the transport branch paulej_logger to bring in the parallel changes from the transport library. In the end, we need to switch the dependencies back to main.


GhostofCookie (approved):

TimEvens (approved):
LGTM

PR #76, opened 2023-09-08
Better gap logging

TimEvens:


include/quicr/gap_check.h
```diff
@@ -32,18 +32,25 @@ namespace quicr {
         std::ostringstream log_s;
 
         if (last_group_id != 0 && last_object_id != 0) {
+            int64_t grp_delta = (group_id - last_group_id);
```
GhostofCookie, line 35: Is that going to convert correctly?

include/quicr/gap_check.h
```diff
@@ -32,18 +32,25 @@ namespace quicr {
         std::ostringstream log_s;
 
         if (last_group_id != 0 && last_object_id != 0) {
```
GhostofCookie, line 34: Could early return and then the ret of the function wouldn't have to be tabbed over so much. But that's not important to this PR.

include/quicr/gap_check.h
GhostofCookie: I suppose it will....

GhostofCookie (approved):

PR #79, opened 2023-09-15
Cleanup

GhostofCookie:
Big scary refactor and cleanup (the formatter was run).

Some key notes:
- Fixed some refs that should have been const&
- Fixed some refs that should have been shared_ptr
- Fixed some shared_ptr that should have been using const& or copy-move
- Fixed some stuttery naming, while using deprecated typealiases for backwards compiling.

Build will probably fail because it relies on https://github.com/Quicr/qname/pull/13 to be merged as well.


include/quicr/gap_check.h
```diff
@@ -1,63 +1,63 @@
 #pragma once
```
GhostofCookie, line 1: This file is a formatting change.

cmd/really/really.cpp
```diff
@@ -137,9 +138,16 @@ class ReallyServer : public quicr::ServerDelegate
...
     qtransport::TransportConfig tcfg{ .tls_cert_filename = "./server-cert.pem",
                                       .tls_key_filename = "./server-key.pem" };
-    server =
-      std::make_unique<quicr::QuicRServer>(relayInfo, tcfg, *this, logger);
+    server = std::make_unique<quicr::Server>(
+      relayInfo, tcfg, shared_from_this(), logger);
+  }
+
+public:
+  static std::shared_ptr<ReallyServer> create()
```
RichLogan, line 146: Couldn't the caller of this just `make_shared` themselves?

GhostofCookie: They could, but follow best practices as per https://en.cppreference.com/w/cpp/memory/enable_shared_from_this

If you use enable_shared_from_this then it's best to not make it constructible to avoid people trying to create non-shared-ptr objects

RichLogan: Okay cool, I figured I was missing something.

cmd/really/really.cpp
```diff
@@ -137,9 +138,16 @@ class ReallyServer : public quicr::ServerDelegate
...
+  {
+    return std::shared_ptr<ReallyServer>(new ReallyServer());
```
bifurcation:
```suggestion
    return std::make_shared<ReallyServer>();
```

GhostofCookie: When this code was still in there that wouldn't have compiled since the constructor was private (see https://en.cppreference.com/w/cpp/memory/enable_shared_from_this). Regardless, this has been removed.

cmd/really/really.cpp
GhostofCookie: Yep, that shared_from_this code was broken, so I removed it, since it would be impossible with that circular dependency setup.

That's why the best practice when using shared_from_this is to make the constructor private, and have a static creation method so that the only way to create the object is as a shared_ptr.

include/quicr/message_buffer.h
```diff
@@ -178,19 +178,18 @@ class MessageBuffer
...
+  template<UnsignedOrName T>
   inline MessageBuffer& operator<<(T value)
   {
     value = swap_bytes(value);
-    value_type* val_ptr = reinterpret_cast<value_type*>(&value);
 
     const auto length = _buffer.size();
     _buffer.resize(length + sizeof(T));
-    std::memcpy(_buffer.data() + length, val_ptr, sizeof(T));
+    std::memcpy(_buffer.data() + length, &value, sizeof(T));
```
bifurcation, line 192: Rather than bothering with endianness, you should just deal with the integral type semantically, and do the big-endian encoding yourself.

```c++
  auto tail = span_type(buffer).subspan(length);
  const auto int_size = tail.size();
  for (int i = 0; i < int_size; i++) {
    tail[i] = static_cast<uint8_t>(value >> unsigned(8 * (int_size - i - 1)));
  }
```

Likewise in `operator>>`

include/quicr/message_types.h
```diff
@@ -29,7 +29,7 @@ enum class MessageType : uint8_t
   PublishIntentEnd,
   Fetch,
 
-  PeerMsg=128
+  PeerMsg = 128
```
bifurcation, line 32: We should be explicit for the other values as well.

include/quicr/quicr_client.h
```diff
@@ -201,7 +202,10 @@ class QuicRClient
                                   bytes&& data);
 
 protected:
-  std::unique_ptr<QuicRClientSession> client_session;
+  std::unique_ptr<ClientSession> client_session;
 };
 
+using QuicRClient
+  [[deprecated("quicr::QuicRClient stutters, use quicr::Client")]] =
```
bifurcation, line 209: Is there a way to move this up a line, or does the annotation need to be right here?

include/quicr/quicr_client_session.h
```diff
@@ -176,4 +179,8 @@ class QuicRClientSession
                                           bytes&& data) = 0;
 };
 
+using QuicRClientSession [[deprecated(
```
bifurcation, line 182: Same comment as above

include/quicr/quicr_common.h
```diff
@@ -1,6 +1,6 @@
 #pragma once
 
-#include <quicr_name>
+#include <qname>
```
bifurcation, line 3: This may be a comment for the `qname` repo and not here, but I would really prefer if non-system headers had a suffix (preferably `.h`) to distinguish them from system headers.

include/quicr/quicr_common.h
```diff
@@ -16,7 +16,7 @@ namespace quicr {
  *    end-to-end, 1400 - 119 = 1281.  A max data size of 1280 should be good
  *    end-to-end for all paths.
  */
-constexpr uint16_t MAX_TRANSPORT_DATA_SIZE = 1200;
+constexpr uint16_t MaxTransportDataSize = 1200;
```
bifurcation, line 19: I generally prefer `snake_case` for values.

include/quicr/quicr_server.h
```diff
@@ -124,7 +123,11 @@ class QuicRServer
                        const messages::PublishDatagram& datagram);
 
 protected:
-  std::unique_ptr<QuicRServerSession> server_session;
+  std::unique_ptr<ServerSession> server_session;
 };
 
+using QuicRServer
+  [[deprecated("quicr::QuicRServer stutters, use quicr::Server")]] =
```
bifurcation, line 130: Same comment as above

include/quicr/quicr_server_session.h
```diff
@@ -109,4 +109,8 @@ class QuicRServerSession
                                const messages::PublishDatagram& datagram) = 0;
 };
 
+using QuicRServerSession [[deprecated(
```
bifurcation, line 112: Same comment as above

src/CMakeLists.txt
```diff
@@ -8,9 +8,9 @@ add_library(quicr
 
 # Suppress external lib warnings
 set_property(GLOBAL PROPERTY RULE_MESSAGES OFF)
-target_compile_options(picotls-core PRIVATE -w)     # TODO Should move to qtransport
-target_compile_options(picotls-openssl PRIVATE -w)  # TODO Should move to qtransport
-target_compile_options(picoquic-core PRIVATE -w)    # TODO Should move to qtransport
+target_compile_options(picotls-core PRIVATE -w)
+target_compile_options(picotls-openssl PRIVATE -w)
+target_compile_options(picoquic-core PRIVATE -w)
```
bifurcation, line 13: Should these no longer move to qtransport?

src/quicr_client.cpp
```diff
@@ -86,19 +90,19 @@ QuicRClient::subscribe(std::shared_ptr<SubscriberDelegate> subscriber_delegate,
 }
 
 void
-QuicRClient::unsubscribe(const quicr::Namespace& quicr_namespace,
-                         const std::string& origin_url,
-                         const std::string& auth_token)
+Client::unsubscribe(const quicr::Namespace& quicr_namespace,
+                    const std::string& origin_url,
+                    const std::string& auth_token)
```
bifurcation, line 95: Are these parameters really needed to unsubscribe?  It's not like you need authorization to leave like you do to join.

src/quicr_client_raw_session.cpp
```diff
@@ -269,50 +263,52 @@ QuicRClientRawSession::on_recv_notify(
...
+                                bytes&& payload,
+                                bool use_reliable_transport)
 {
 
   if (pub_delegates.count(quicr_namespace)) {
     return true;
   }
 
-  pub_delegates[quicr_namespace] = pub_delegate;
+  pub_delegates[quicr_namespace] = std::move(pub_delegate);
```
bifurcation, line 278: Generally, prefer `insert_or_assign` to using `operator[]` and `=`.

bifurcation: Similar comments below.

src/quicr_server_raw_session.cpp
```diff
@@ -107,14 +105,13 @@ QuicRServerRawSession::run()
...
-  const quicr::Namespace& quicr_namespace,
-  const PublishIntentResult& result)
+ServerRawSession::publishIntentResponse(const quicr::Namespace& quicr_namespace,
+                                        const PublishIntentResult& result)
 {
   if (!publish_namespaces.count(quicr_namespace))
     return;
 
-  // TODO: Need to update publish_namespaces and intent methods to support multi origin
+  // TODO: Support more than one publisher per ns
```
bifurcation, line 114: This actually seems like something we should not support, to avoid sending of duplicate names.

src/quicr_server_raw_session.cpp
```diff
@@ -389,28 +384,29 @@ QuicRServerRawSession::TransportDelegate::on_connection_status(
...
     }
 
-    for (auto &ns: pub_names_to_remove) {
+    for (auto& ns : pub_names_to_remove) {
       server.publish_namespaces.erase(ns);
     }
 
     std::vector<quicr::Namespace> sub_names_to_remove;
-    for (auto& [ns, sub_map]: server.subscribe_state) {
+    for (auto& [ns, sub_map] : server.subscribe_state) {
```
bifurcation, line 405: Delete extra blank line below this one.

bifurcation (approved):
Approved with nits fixed.

include/quicr/message_types.h
GhostofCookie: This is a hack that came later, it's the only one that specifically needs that number. The rest do not.

include/quicr/quicr_client.h
GhostofCookie: You can thank clang-format and punch cards here. Attributes for using statements have to come before the =, so unfortunately it looks like this. Thanks punch cards!

include/quicr/quicr_common.h
GhostofCookie: Me too. I've had push back for snake case in this repo but I'm happy to rename this one. I'd rename it all if I could

include/quicr/message_buffer.h
GhostofCookie: We are doing it ourselves in swap_bytes. And it's compile time evaluated, so it doesn't cost anything to check endianness.

include/quicr/message_buffer.h
```diff
@@ -178,19 +178,18 @@ class MessageBuffer
...
-   * @tparam T A type satisfying quicr::is_integral.
+   * @tparam T An unsigned integral or a quicr::Name.
    * @param value The message to be written.
    * @returns The MessageBuffer that was written to.
    */
-  template<typename T, typename = std::enable_if_t<quicr::is_integral_v<T>, T>>
+  template<UnsignedOrName T>
   inline MessageBuffer& operator<<(T value)
   {
     value = swap_bytes(value);
```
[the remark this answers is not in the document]
GhostofCookie: Not going to lie, it took me longer than I'm proud of to understand what to_be even meant. Just seemed like unnecessary misdirection for second ("value is to be value").

src/CMakeLists.txt
GhostofCookie: Turns out it has to be in each repo :) Not just transport.

src/quicr_client.cpp
GhostofCookie: No idea. But I'm avoiding crazy API changes as much as possible for now. Would love to address this in a follow up PR though.

cmd/really/reallyTest.cpp
```diff
@@ -127,7 +127,7 @@ main(int argc, char* argv[])
 
   qtransport::TransportConfig tcfg{ .tls_cert_filename = NULL,
                                     .tls_key_filename = NULL };
-  quicr::QuicRClient client(relay, tcfg, logger);
+  quicr::Client client(relay, tcfg, logger);
```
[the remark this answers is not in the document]
GhostofCookie: Seems like an abuse of auto.... any particular reason for this?

src/quicr_server_raw_session.cpp
GhostofCookie: Leaving this unresolved and adding @TimEvens to the conversation.

PR #80, opened 2023-09-21
Fixed fragments being shared between client objects.

GhostofCookie:


src/quicr_client_raw_session.cpp
```diff
@@ -623,22 +600,23 @@ QuicRClientRawSession::handle_pub_fragment(
...
+  auto& curr_fragments = fragments[cindex];
+  const auto& msg_iter = curr_fragments.find(datagram.header.name);
+  if (msg_iter != curr_fragments.end()) {
     // Found
     auto& [_, buffer] = *msg_iter;
     buffer.emplace(datagram.header.offset_and_fin,
                    std::move(datagram.media_data));
     if (notify_pub_fragment(datagram, delegate, buffer))
-      fragments[cindex].erase(msg_iter);
+      curr_fragments.erase(msg_iter);
```
bifurcation, line 611: Surround this line with `{ }`

bifurcation (changes requested):
Thanks @GhostofCookie, this looks good to me with the one small issue noted fixed.

In the spirit of [Chesterton's fence](https://en.wikipedia.org/wiki/G._K._Chesterton#Chesterton's_fence), I'm trying to imagine how a `static` reassembly buffer ever made sense.  I guess you could have a shared fragment cache so that multiple clients accessing the same content could share fragments.  But that doesn't appear to be what was being done here.

The change will impose more overhead when running multiple clients, but that seems acceptable compared to the complexity of having a real shared fragment cache.

bifurcation (approved):

PR #81, opened 2023-09-22
Apply clang-tidy

bifurcation:
This PR brings in enforcement of clang-tidy.  It updates the code in the repo so that clang-tidy reports no errors with the relatively conservative `.clang-tidy` file that is checked in here.  It also adds clang-tidy to the CI builds so that they will enforce conformance for future PRs.  Also, for some reason, `make format` wasn't covering `cmd`, so I added that.

Things for folks to look at in review:
* Check that the changes to using `.empty()` and `.contains()` don't accidentally reverse polarity
* I highlighted one spot with `XXX(richbarn)` comments where the answer wasn't clear to me (and one where the test is just wrong)

`clang-tidy` is currently disabled for benchmarks, because it seemed difficult to exclude of all the dependency code that is dynamically pulled in.  But it would be good if the benchmark code in this repo could be checked in a future PR.


src/quicr_client_raw_session.cpp
```diff
@@ -492,11 +502,13 @@ ClientRawSession::publishNamedObject(const quicr::Name& quicr_name,
     }
 
     // Send last fragment, which will be less than max_transport_data_size
-    if (frag_remaining_bytes) {
+    if (frag_remaining_bytes > 0) {
       messages::MessageBuffer msg;
+      // NOLINTNEXTLINE(hicpp-signed-bitwise)
```
GhostofCookie, line 507: Interesting... does clang-tidy shut up if the code is changed to
```cpp
datagram.header.offset_and_fin = uintVar_t((offset << 1u) + 1u);
```
?

src/quicr_client_raw_session.h
```diff
@@ -275,7 +276,10 @@ class ClientRawSession
   bool has_shared_transport{ false };
   std::atomic_bool stopping{ false };
   qtransport::StreamId transport_dgram_stream_id{ 0 };
-  qtransport::TransportContextId transport_context_id;
+  // XXX(richbarn): I added an initializer for this variable for this member
+  // that's of the same form as the ones above.  But it's not clear to me that
+  // this is actually the right answer.
+  qtransport::TransportContextId transport_context_id{ 0 };
```
GhostofCookie, line 282: It may be, but then some added checking should probably go into the constructor or the `connect` method:
```
if context_id == 0
  throw We haven't started transport! Panic!!!

File: src/quicr_server_raw_session.cpp [hunk 3]
```diff
@@ -60,7 +60,7 @@ ServerRawSession::ServerRawSession(
   : delegate(std::move(delegate_in))
   , logger(std::make_shared<cantina::Logger>("QSES", logger))
   , transport_delegate(*this)
-  , transport(transport_in)
+  , transport(std::move(transport_in))
```
GhostofCookie, line 63: Nice catch!

src/quicr_server_raw_session.cpp
```diff
@@ -324,12 +320,11 @@ ServerRawSession::handle_publish_intent(
...
-        // TODO: Resend response?
-        break;
+      // TODO(trigaux): Already registered this namespace successfully, do
+      // nothing?
       case PublishIntentContext::State::Ready:
-        // TODO: Already registered this namespace successfully, do nothing?
-        break;
       default:
         break;
     }
```
GhostofCookie, line 330: Surprised compiler didn't complain about fallthrough:
```cpp
      case PublishIntentContext::State::Pending:
        [[fallthrough]] // TODO: Resend response?
      case PublishIntentContext::State::Ready:
        [[fallthrough]] // TODO: Already registered this namespace successfully, do nothing?
      default:
        break;

File: test/encode.cpp [hunk 5]
```diff
@@ -15,23 +15,26 @@ using namespace quicr::messages;
...
+  const auto u32 = uint32_t(0x12345678U);
+  const auto u64 = uint64_t(0x123456789ABCDEF0U);
+  const auto u128 = 0x123456789ABCDEF0123456789ABCDEF0_name;
+
+  // XXX(richbarn): This test will return incorrect results on a big-endian
+  // system, since swap_bytes actually converts to big-endian.
+  CHECK(u16 != swap_bytes(u16));
+  CHECK(u32 != swap_bytes(u32));
+  CHECK(u64 != swap_bytes(u64));
+  CHECK(u128 != swap_bytes(u128));
```
GhostofCookie, line 28:
```suggestion
  if constexpr (std::endian::native == std::endian::little) {
    CHECK(u16 != swap_bytes(u16));
    CHECK(u32 != swap_bytes(u32));
    CHECK(u64 != swap_bytes(u64));
    CHECK(u128 != swap_bytes(u128));
  }
```

GhostofCookie: What's wrong with `CHECK_NE`?

src/quicr_client_raw_session.h
bifurcation: Looking at the code, it seems that this field is [not actually initialized until `connect()`](https://github.com/Quicr/libquicr/blob/main/src/quicr_client_raw_session.cpp#L92C45-L92C45).  So I would suggest changing its type to `std::optional<qtransport::TransportContextId>`.  That will make the initialization error go away, and will force other methods to check that it is present (and thus the transport is connected) before using it.  (As opposed to just [blindly assuming it is initialized](https://github.com/Quicr/libquicr/blob/main/src/quicr_client_raw_session.cpp#L260).)  It's the safer, clearer way of doing your `if context_id == 0` trick.

src/quicr_server_raw_session.cpp
bifurcation: Thanks, clang-tidy!

test/encode.cpp
bifurcation: I think there was something internal to it that clang-tidy was unhappy with, but I also just think that `_EQ` / `_NE` suffixes are less clear than just writing something like you would in an `if` statement.  Note that doctest rewrites the equality operators so that they do the right thing for it.

cmd/really/reallyTest.cpp
```diff
@@ -105,40 +104,44 @@ main(int argc, char* argv[])
     std::cerr << std::endl;
     std::cerr << "Usage PUB: reallyTest FF0001 pubData" << std::endl;
     std::cerr << "Usage SUB: reallyTest FF0000" << std::endl;
-    exit(-1);
+    exit(-1); // NOLINT(concurrency-mt-unsafe)
```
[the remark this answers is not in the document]
bifurcation: Only for things that are unsafe in concurrent envrionments.

src/quicr_client_raw_session.cpp
bifurcation: Yep, good call.  (It wants `U`, not `u`, but whatever.)

src/quicr_server_raw_session.cpp
bifurcation: clang-tidy complained about identical branches, which it still does even with the `[[fallthrough]]` annotation.  But I `NOLINT`'ed it.

test/encode.cpp
bifurcation: Actually, I just went ahead and fixed the test.  I created byte-swapped values for the little-endian case, and compared the output of `swap_bytes` to the correct value for each endianness.

paulej:
If you want to incorporate that into the local build process, something like this is what I use:

```cmake
# Locate clang-tidy
find_program(CLANG_TIDY_EXE NAMES "clang-tidy" REQUIRED)

# Set the clang-tidy command with a list of checks
list(APPEND TIDY_OPS "-modernize-use-trailing-return-type")
list(APPEND TIDY_OPS "cppcoreguidelines-*")
list(APPEND TIDY_OPS "-cppcoreguidelines-pro-type-union-access")
list(APPEND TIDY_OPS "-cppcoreguidelines-pro-type-reinterpret-cast")
list(APPEND TIDY_OPS "-cppcoreguidelines-owning-memory")
list(APPEND TIDY_OPS "-cppcoreguidelines-pro-bounds-pointer-arithmetic")
list(APPEND TIDY_OPS "-cppcoreguidelines-avoid-magic-numbers")

set(CLANG_TIDY_COMMAND "${CLANG_TIDY_EXE}" "-checks=modernize-*")
foreach(TIDY_OP IN LISTS TIDY_OPS)
    set(CLANG_TIDY_COMMAND "${CLANG_TIDY_COMMAND},${TIDY_OP}")
endforeach()

# Specify the C++ standard to observe
set_target_properties(target_name_here
    PROPERTIES
        CXX_STANDARD 20
        CXX_STANDARD_REQUIRED ON
        CXX_EXTENSIONS OFF
        CXX_CLANG_TIDY "${CLANG_TIDY_COMMAND}")
```

bifurcation:
@paulej You can actually put all that customization in a `.clang-tidy` file, which is what this PR does.

src/encode.cpp
```diff
@@ -391,7 +391,7 @@ operator<<(MessageBuffer& buffer, const Header& msg)
   return buffer;
 }
 
-MessageBuffer&
+static MessageBuffer&
```
GhostofCookie, line 394: Why just these 2?

bifurcation: Because those are the two for which clang-tidy complained about missing declarations.  If you want, we could declare them in `encode.h`, but it didn't seem necessary to expose them.

cmd/really/really.cpp
```diff
@@ -141,38 +142,36 @@ class ReallyServerDelegate : public quicr::ServerDelegate
...
-    [[maybe_unused]] bool use_reliable_transport,
-    quicr::messages::PublishDatagram&& datagram)
+  void onPublisherObject(const qtransport::TransportContextId& context_id,
+                         [[maybe_unused]] const qtransport::StreamId& stream_id,
+                         [[maybe_unused]] bool use_reliable_transport,
+                         quicr::messages::PublishDatagram&& datagram) override
   {
-    std::list<Subscriptions::Remote> list =
-      subscribeList.find(datagram.header.name);
+    const auto list = subscribeList.find(datagram.header.name);
```
paulej, line 172: IMO, use of auto is "bad". The keyword IMO is a somewhat bad introduction in C++, anyway. I have abused it where I was too lazy to specify the type, but I definitely would not purposely change the code like this.  If there is a type change, this would catch it.  That might be good or bad, but I would argue it is generally better to be explicit.  One must use `auto` on lamdas and there are some functions where `auto` is necessary since the input type might be a plurality of things.  So while there are valid uses for it, I would limit use to those cases where it's truly the best choice.

cmd/really/subscription.cpp
```diff
@@ -1,59 +1,63 @@
...
-  else {
-    std::set<Remote>& list = mapPtr->second;
-    if ( list.find( remote ) == list.end() ) {
-      list.insert( remote ); // TODO - rethink if list is right thing here 
+    list.insert(remote);
+    std::pair<quicr::Name, std::set<Remote>> pair;
+    pair = make_pair(prefix.name(), list);
+    subscriptions[len].insert(pair);
+  } else {
+    auto& list = mapPtr->second;
```
paulej, line 22: Here is an example of where `auto` made the code harder to read. Whereas one would have previously known the type, now one has to go digging into a map to figure it out.

src/encode.cpp
```diff
@@ -63,16 +67,14 @@ create_transaction_id()
...
 }
 
 MessageBuffer&
 operator>>(MessageBuffer& msg, Namespace& value)
 {
-  Name name;
-  uint8_t length;
-  // TODO: These should be swapped around.
+  auto name = Name{};
+  auto length = uint8_t(0);
```
paulej, line 77: `uint8_t length{}` would address the initialization issue and avoids `auto`.

src/message_buffer.cpp
```diff
@@ -68,46 +66,53 @@ MessageBuffer::push(buffer_type&& data)
...
+  }
 
   cleanup(length);
 };
 
 const uint8_t&
 MessageBuffer::front() const
 {
-  if (empty())
+  if (empty()) {
```
paulej, line 79: These braces are better here, though I think the root of the problem is the fact that the clang-format file creates this styling.  What I personally prefer is to keep short `if` statements on a single line.  The syntax I use for that is `AllowShortIfStatementsOnASingleLine: WithoutElse`.

src/quicr_client_raw_session.cpp
```diff
@@ -468,13 +491,14 @@ ClientRawSession::publishNamedObject(const quicr::Name& quicr_name,
       /*
        * For UDP based transports, some level of pacing is required to prevent
        * buffer overruns throughput the network path and with the remote end.
-       *  TODO: Fix... This is set a bit high because the server code is running
-       * too slow
+       *  TODO(paulej): Fix... This is set a bit high because the server code is
```
paulej, line 489: Not my TODO. My name appeared here only because I moved the code into this file.

src/quicr_server_raw_session.cpp
```diff
@@ -323,16 +319,16 @@ ServerRawSession::handle_publish_intent(
     publish_namespaces[intent.quicr_namespace] = context;
   } else {
     auto state = publish_namespaces[intent.quicr_namespace].state;
+    // NOLINTBEGIN(bugprone-branch-clone)
     switch (state) {
       case PublishIntentContext::State::Pending:
-        // TODO: Resend response?
-        break;
+        [[fallthrough]]; // TODO(trigaux): Resend response?
```
paulej, line 325: I assume the intent is to do something here and not actually fall through?

cmd/really/really.cpp
bifurcation: It's not about laziness.  See [Almost Always Auto](https://herbsutter.com/2013/08/12/gotw-94-solution-aaa-style-almost-always-auto/)

cmd/really/subscription.cpp
bifurcation: And when we refactor to change the type of the map, we'll have to go change a dozen unnecessary type declarations.

src/message_buffer.cpp
bifurcation: Strong disagree.  Too much risk of [`goto fail`](https://www.synopsys.com/blogs/software-security/understanding-apple-goto-fail-vulnerability-2.html) and related bugs.  Note that `AllowShortIfStatementsOnASingleLine: WithoutElse` would not have caught `goto fail`.  Security risk trumps aesthetics.

src/quicr_server_raw_session.cpp
bifurcation: The `[[fallthrough]]` just consolidates the `break` statements.  No change to behavior intended, until we resolve the `TODO`s here.

cmd/really/really.cpp
paulej: This is like taking a step toward loose typing. I totally disagree. I could write a blog post on why `auto` should be considered harmful. It's just bad. It makes sense for some situations, but it most definitely should not be used almost always.

cmd/really/subscription.cpp
paulej: Yeah! And, good! What is "second" here? I cannot tell with that `auto` there.  If we change the type of whatever `.second` is, we should be forced to inspect downstream impacts.

cmd/really/really.cpp
bifurcation: Every modern programming language disagrees with you, even the ones with very strong type systems (some far stronger than C++).  Go has `:=`, Rust has `let`, Swift has `let` and `var`, Kotlin has `val` and `var`, ... even Java has `var`.  Type inference is the modern way.

src/message_buffer.cpp
paulej: The only risk is when `if` requires more than one line. In this `goto` example, if the `goto` was on the same line as `if`, people would not be adding an extra `goto` lines. The problem is when the code starts looking like Python -- as it did in this case. People then feel compelled to add more lines.

cmd/really/really.cpp
paulej: I think we all need to start writing in Ada.

paulej: More seriously, though, `auto foo = something->second` just makes reading code harder since one has to go hunt down what `second` is. I've run into the same thing in Python.  It's just bad for managing code. It works, but bad. I don't care if other languages want to create the same bad practice.

suhasHere (approved):
thanks for the PR Richard. 

Looks good

PR #82, opened 2023-09-22
Make circular index a member variable

bifurcation:
This PR makes the circular index `c_index` a member variable instead of a static variable in a method.  This ensures that different instances don't interfere with each other.

... at least that was how this PR started.  Once I got in there, I noticed that the fragmentation buffer logic was unnecessarily complicated, with a bunch of unnecessary (and unsafe) iterator logic, unnecessary `std::map` where `std::array` would work fine, and unnecessary duplication across branches.  So I rewrote the whole method in  way that should make it simpler, faster, and safer.


paulej (approved):
Changes look good, though I'd prefer we use the Frame Assembler object from the src/h3/frame_assember.[cpp|h] in place of the assembler code in this module since I think it's provides for a good separation of logic.  That code is in the paulej_h3 branch.

src/quicr_client_raw_session.h
```diff
@@ -278,29 +278,32 @@ class ClientRawSession
...
+  // executed. If not, it'll only update the map with the new offset value.
+  // Incomplete messages can exist in the cache for as long as the circular
+  // index hasn't wrapped to the same point in cache.  Under high
+  // load/volume, this can wrap within a minute or two.  Under very little
+  // load, this could linger for hours. This is okay considering the only harm
+  // is a little extra memory being used. Extra memory is a trade-off for being
+  // event/message driven instead of timer based with threading/locking/...
+  uint32_t circular_index{ 0 };
+  static constexpr size_t max_pending_per_buffer = 5000;
+  static constexpr size_t max_fragment_buffers = 20;
```
TimEvens, line 303: This should be configurable. 20 only allows for a little over 24k. Right? We have messages at least up to 80k.  The design was to allow up to 10MB, although we don't go that large.

TimEvens (commented):
Array is. I've but size is fixed. Can you make that dynamic or at least configurable?

suhasHere (approved):
looks fine to me. Also agree with Tim, we might want to up the constants

src/quicr_client_raw_session.h
bifurcation: What sort of configurable are you thinking of?  Right now, it's compile-time configurable through these parameters, just like it was before.  Also these constants are the same as before.  I hope you don't mean we need runtime configurability?

The buffer capacity is a little deceptive here.  The `max_pending_per_buffer` is in units of messages; the amount of data in each message is unbounded.  (Maybe it shouldn't be!)  So you can have 10MB messages, as long as you don't have more than 100k in flight at once.  That comes out to 1TBps, so I think we're probably OK :)

GhostofCookie: I think the thought was that we could maybe make this runtime configurable from the client maybe?

GhostofCookie (approved):
Nothing I disapprove of here, though I would agree with Paul that we should use a FragmentAssembler object. But that can be for a separate PR. And I agree with Tim that runtime configurable would be nice as well.

src/quicr_client_raw_session.h
bifurcation: I don't think we should do that.  To avoid making this massively complex, you would have to make sure that the client only set these values once, and given that we've got plenty of capacity, it doesn't seem worth even the modest complexity of making set-once values.

bifurcation:
I'll leave runtime configurability and separate objects for follow-ons.

src/quicr_client_raw_session.h
TimEvens: Right, a little hard to see it on my phone. IMO it shouldn't be unbounded. It should be configurable. I was reading the two the other way around, so yeah we have 5000 fragments with only 20 (fixed) in flight. It works for now... but considering all the effort for a refactor it's a bit of a miss tuning wasn't added for runtime. Also, it's a bit of a miss that the refactor didn't include fixing the fixed 1200 byte fragments.

PR #84, opened 2023-10-05
Remove circular Server Delegate dependency

GhostofCookie:
A bit of an API change, this affects server only. The work here was to address the TODOs in really (which in effect affects LAPS) which were to remove the circular dependency between the server and server delegate. Basically, the changes just pull any calls to server methods out of the delegate, so that mechanics of responding are not the responsibility of the application.

### DIFF
--- a/cmd/really/really.cpp
+++ b/cmd/really/really.cpp
@@ -133,15 +133,6 @@ public:
   {
   }
 
-  /**
-   * Hacky dependency injection.
-   * TODO(trigaux): Remove this once delegate no longer depends on server.
-   */
-  void setServer(std::shared_ptr<quicr::Server> server_in)
-  {
-    server = std::move(server_in);
-  }
-
   quicr::PublishIntentResult onPublishIntent(
     const quicr::Namespace& quicr_namespace,
     const std::string& /* origin_url */,
@@ -168,8 +159,6 @@ public:
     quicr::messages::PublishDatagram&& datagram) override
   {
     auto list = subscribeList.find(datagram.header.name);
-    std::erase_if(list,
-                  [&](const auto& s) { return s.context_id == context_id; });
 
     std::vector<quicr::PublishResult> matching_subscriptions;
     for (const auto remote : list) {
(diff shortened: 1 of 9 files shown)


cmd/really/reallyTest.cpp
```diff
@@ -104,28 +102,27 @@ main(int argc, char* argv[])
...
     std::cerr << "Usage PUB: reallyTest FF0001 pubData" << std::endl;
     std::cerr << "Usage SUB: reallyTest FF0000" << std::endl;
-    exit(-1); // NOLINT(concurrency-mt-unsafe)
   }
 
   // NOLINTNEXTLINE(concurrency-mt-unsafe)
-  const auto* relayName = getenv("REALLY_RELAY");
-  if (relayName != nullptr) {
+  const char* relayName = getenv("REALLY_RELAY");
+  if (!relayName) {
```
GhostofCookie, line 109: This was a segfaulting bug.

include/quicr/quicr_common.h
```diff
@@ -76,6 +76,13 @@ struct RelayInfo
   Protocol proto;       // Transport protocol to use
 };
 
+struct PublishResult
+{
+  std::uint64_t subscription_id;
+  std::uint8_t priority;
+  std::uint16_t expiry_ms;
+};
```
GhostofCookie, line 84: This probably needs some love, just got something simple to start.

cmd/really/really.cpp
```diff
@@ -164,23 +152,24 @@ class ReallyServerDelegate : public quicr::ServerDelegate
...
-      if (dest.context_id == context_id) {
-        // split horizon - drop packets back to the source that originated the
-        // published object
-        continue;
-      }
-
-      // TODO(trigaux): Move logic into quicr::Server
-      server->sendNamedObject(dest.subscribe_id, false, 1, 200, datagram);
+    auto list = subscribeList.find(datagram.header.name);
+
```
GhostofCookie, line 162: I had added dedup here, but after testing with Decimus I removed it as it only really hindered testing. Joining a Team meeting is fine, you won't receive yourself because you didn't subscribe to it. Mirror calls work very nicely.

PR #86, opened 2023-10-10
Bump transport version

RichLogan:
It seems the transport was inadvertently(?) downgraded in https://github.com/Quicr/libquicr/pull/85, reving to latest ([8123f13](https://github.com/Quicr/transport/commit/8123f132685da708a17af8e05093ad1b6d78691d))


bifurcation:
Definitely inadvertent, sorry.

PR #89, opened 2023-10-27
Support priority pass down to transport

TimEvens:
* Updates publishIntent API to add priority, which is passed to transport


GhostofCookie (approved):
Build failing, but updating qname should fix that

TimEvens:
> Build failing, but updating qname should fix that

Thanks! Just updated qname submodule

TimEvens:
Hitting https://github.com/actions/runner-images/issues/8659 issue.  Switching ubuntu build to ubuntu-20.04 since we do not require 22.04.

PR #101, opened 2023-11-20
Fix disconnect not disconnecting.

GhostofCookie:


src/quicr_client_raw_session.cpp
```diff
@@ -90,20 +92,27 @@ ClientRawSession::~ClientRawSession()
 bool
 ClientRawSession::connect()
 {
+  if (!transport) {
+    throw std::runtime_error(
+      "Transport has been destroyed, create a new session object!");
+    return false;
```
TimEvens, line 98: Throw and return?

TimEvens (approved):
LGTM, just one comment on throw/return.

src/quicr_client_raw_session.cpp
GhostofCookie: Oops!

PR #127, opened 2024-02-14
Add reset and wait transport feature

TimEvens:


src/quicr_client_raw_session.cpp
```diff
@@ -521,7 +526,6 @@ ClientRawSession::publishNamedObject(const quicr::Name& quicr_name,
       if (transport->enqueue(context.transport_conn_id, context.transport_data_ctx_id, msg.take(),
                              priority, expiry_age_ms, eflags) !=
           qtransport::TransportError::None) {
-        std::this_thread::sleep_for(std::chrono::microseconds(100));
```
RichLogan, line 524: I don't actually know why this was here to begin with but assuming this removal is fine.

RichLogan (approved):

src/quicr_client_raw_session.cpp
TimEvens: It was there when we used safe_queue and would be dropping messages. The sleep isn't needed anymore.

PR #128, opened 2024-02-15
Add pause/resume feature support

### LINKED ISSUE #126: Implement subscriber resume/pause message and relay support
Decimus will use a new Qmedia API to instruct stopping or starting of subscriptions. 

Transport mode `Pause` and `Resume` will be added to instruct the relay to stop/start sending data for a subscription.  

Pause/Resume is only valid for existing subscription that has a valid `subscribe()`  with the desired transport mode as reliable/unreliable/...  Pause/Resume is just to start/stop data flows for the subscription and will not be propagated via peers/etc.  Qmedia will send  `subscribe()` with transport mode to `Pause` to stop flows. Qmedia will start data flow using `subscribe()` with `Resume` transport mode.  Qmedia will send a subscribe based on start/stop to libquicr.

TimEvens:
Resolves #126


RichLogan:
I was going to add some tests to this but looks like we're missing some infrastructure to make that straight forward.

RichLogan (approved):

PR #131, opened 2024-02-26
Allow state update for existing subscriptions

RichLogan:


TimEvens (approved):
Looks good. I'll follow up with another PR to update it a little with what we talked about.

PR #152, opened 2024-04-17
Fix compare and add chunk_size to client constructor

TimEvens:
Fixes a compare build error and adds chunk size to quicr client constructor. This allows clients to configure their stream chunk (aka fragment) size.  Zero will disable the feature and will not chunk/fragment.

### DIFF
--- a/cmd/really/reallyTest.cpp
+++ b/cmd/really/reallyTest.cpp
@@ -228,7 +228,7 @@ main(int argc, char* argv[])
     .tls_key_filename = nullptr,
   };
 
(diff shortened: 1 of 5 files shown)


suhasHere (approved):

RichLogan:
Should we add a comment to the method doc to say that 0 disables, and that if it is set it the actual value will be `max(chunk_size, max_transport_data_size)`?

RichLogan (approved):

TimEvens:
> Should we add a comment to the method doc to say that 0 disables, and that if it is set it the actual value will be `max(chunk_size, max_transport_data_size)`?

I'll update the comment.

PR #155, opened 2024-04-29
Add missing metrics macro

RichLogan:


TimEvens:
@RichLogan , apparently GitHub made a change to their macOS runners and go is no longer installed. I updated that in the previous #154. The main branch now has this build test should not fail on missing go.

TimEvens (approved):
Looks good, but there are more places now that would need to be updated. This will hopefully be moot very soon as we move metrics over to m10x pub/sub instead of needing influxdb-cxx/cpr/libquicr.

TimEvens:
No longer needed and metrics change to use m10x metrics over pub/sub replaces this.

PR #159, opened 2024-05-15
Update transport to main branch and fix fragment ooo handling

TimEvens:


src/quicr_client_raw_session.cpp
```diff
@@ -759,19 +759,12 @@ ClientRawSession::notify_pub_fragment(
     }
 
     reassembled.insert(reassembled.end(),
-                       std::make_move_iterator(data.begin()),
-                       std::make_move_iterator(data.end()));
+                       data.begin(),
+                       data.end());
```
GhostofCookie, line 763: Are you sure you don't want to move the data?

GhostofCookie (approved):

PR #167, opened 2024-05-31
Publish metrics

GhostofCookie:
Adding measurements API to submit metrics via pub/sub over quic.


src/quicr_client_raw_session.cpp
```diff
@@ -218,6 +241,119 @@ ClientRawSession::connecting() const
...
+  const std::string measurement_str = measurement_json.dump();
+  quicr::bytes measurement_bytes(measurement_str.begin(), measurement_str.end());
+
+  const auto start_time = std::chrono::time_point_cast<std::chrono::microseconds>(std::chrono::steady_clock::now());
+  std::vector<qtransport::MethodTraceItem> trace;
+  trace.reserve(10);
+  trace.push_back({"libquicr:publishMetrics:begin", start_time});
+
+  publishNamedObject(*metrics_namespace, 31, 500, std::move(measurement_bytes), std::move(trace));
+}
```
suhasHere, line 355: looks like we are publishing without subscription ?
Also we are not going through intent before publishing ..

GhostofCookie: Metrics adapter is always subscribed to the server (so long as it's running).

We send a publish intent when we receive a connection response. When we receive a publish intent response we start the running the metrics thread.

As for sending individual metrics through the API, this follows the same logic as publishNamedObject; You can try publishing without an intent, but it won't have any effect if you haven't sent a publish intent first.

suhasHere:
Thanks ..  I will review this PR later today or tomorrow more thoroughly

TimEvens (approved):
LGTM for now. Upon merge, it will break things, but I will fix that with an update to MOQT.

PR #170, opened 2024-06-04
Add reliable flag to server publish object cb

TimEvens:
Relay needs to know if the received object was
via a reliable stream or datagram.


shenning00 (approved):
Approved. Thanks for updating.

PR #171, opened 2024-06-14
Added optional build with mbedtls over boringssl.

### LINKED ISSUE #162: epic: Integrate mbedTLS using picoquic support via picoTLS
Update libquicr and submodules to support mbedTLS over OpenSSL. 

* Picoquic supports mbedTLS via picotls.  Update transport to configure picoquic/picotls to use mbedTLS instead of boringSSL.
* Configure other submodules to support/use mbedtls instead of openssl. Make sure openssl is not required or used by submodules. If they are, use mbedtls via the transport in the same way that boringssl is used today.

BrettRegnier:
Additionally: 'Fixed' compile error saying that fake_transport function setDataCtxPriority doesn't override anything.

Note. This should be merged into main first, and the submodule file updated accordingly.
https://github.com/Quicr/transport/pull/149

Closes #162


TimEvens (approved):
LGTM, I'll double check and then merge.

RichLogan (changes requested):
Just noting this here so we don't merge it broken. 

This doesn't work for me out of the box as the python dependency install in transport fails. I think I have a fix in transport though that should work in all cases: https://github.com/Quicr/transport/pull/154

As an aside, should we add a CI variant for an mbedtls build?

RichLogan (approved):

TimEvens:
Transport and this branch still depended on the host system having mbedtls and openssl installed. The latest changes resolves it and it builds now in docker where the host/container does not have openssl or mbedtls installed.  

This should be ready to merge.

PR #174, opened 2024-06-27
Implement draft-ietf-moq-transport-04

### LINKED ISSUE #136: Add MoQ message encode/decode.
MoQ message needs encoder and decoder support.

There is start of this on this dev branch: https://github.com/Quicr/libquicr/blob/moq-suhas-feb2024/src/moq_message.cpp 

We should see if it makes sense to use it as-is or rewrite as needed

### LINKED ISSUE #137: QuicrClient: Support Message Handling for MoQ Mesages
This work requires few changes
1: Mapping between MOQ identifiers ( track-alias, subscribe-id, full trackname) to QUICR identifiers ( Namespace, Name)
2. Support handlers for MoQ Messages
3. Handle subscriptions ( for publishers) forwarded
4. URI convertor integration

### LINKED ISSUE #161: MOQT: Implement moqt protocol state machine
Change the current `libquicr` implementation to be compliant with MOQT protocol state machine. 

Most of libquicr works in a similar fashion to MOQT with the following changes:

* Add additional connection parameters
* Track alias needs to be created and conveyed via subscribe, server will accept (always) or deny with subscribe error
* Publisher does not start sending after announce (aka publish intent). It waits till it receives a subscribe from the relay.
* Consider sending announce cancel long with unannounce
* Publisher matching will only be on namespace, not on track name. When the relay receives data from publisher, it will publish with a track alias that was created based on a matching received subscribe that includes both track namespace and name. The first and subsequent subscribes will be associated to this publisher track alias, enabling received objects with that track alias to be forwarded to all subscribers of the same track namespace/track name. 
* Publisher will not be able to send other tracks till it gets a subscribe from the relay for a specific track. Relay has no idea what to subscribe to (e.g., the relay knows nothing about namespaces or track names), therefore the relay requires some other client to first subscribe for it. 
* When subscribers are zero (similar to a reference count for a given object), the relay will unsubscribe

### LINKED ISSUE #173: Implement draft-ietf-moq-transport-04
Implement all control messages and all object data stream modes.  

Refactor libquicr quicr client/server into moq instance that can be used by both server and client. Refactor API to be track based. Abstract in the API much of the work and state that MOQT requires. The client should have simple subscribe and publish track methods to establish data flows. 

Ensure metrics continue to work with MOQ implementation. 

Create new mod-example client/server programs to show how to use the client APIs for both client and server.  Include a date publishing/subscribe implementation in the example. 

Update the README for MOQ.

TimEvens:
Implements draft-ietf-moq-transport-04 in parallel with existing
Libquicr client/server. QuicrClient/Server and MOQIntsance can be used
side by side.   New `mod-example` program replaces `really`. 

* Resolves #136
* Resolves #137
* Resolves #161
* Resolves #173


cmd/moq-example/client.cpp
```diff
@@ -0,0 +1,396 @@
...
+            _logger->info << "Send message: " << msg << std::flush;
+        }
+
+        if (object_id % 5 == 0) {       // Set new group
+            object_id = 0;
+            group_id++;
+        }
+
+        std::vector<uint8_t> msg_v(msg.begin(), msg.end());
+        auto result = track_delegate->sendObject(group_id, object_id++, msg_v);
```
GhostofCookie, line 184: It is possible to avoid copying the string here by interpreting the string as a byte*:
```cpp
auto result = track_delegate->sendObject(group_id, object_id++, std::span{std::bit_cast<uint8_t*>(&msg), sizeof(std::string)});
```
Probably a good idea to check that cast, but that's the general idea.

src/moq_instance.cpp
```diff
@@ -0,0 +1,1687 @@
...
+
+        announce.track_namespace.assign(track_namespace.begin(), track_namespace.end());
+        announce.params = {};
+        buffer <<  announce;
+
+        std::vector<uint8_t> net_data = buffer.front(buffer.size());
+
+        _logger->debug << "Sending ANNOUNCE to conn_id: " << conn_ctx.conn_id << std::flush;
+
+        send_ctrl_msg(conn_ctx, buffer.front(buffer.size()));
```
GhostofCookie, line 189: Net_data isn't used, should it be? Looks like it uses the same bytes.

src/moq_instance.cpp
```diff
@@ -0,0 +1,1687 @@
...
+        auto announce_ok = MoqAnnounceOk{};
+
+        announce_ok.track_namespace.assign(track_namespace.begin(), track_namespace.end());
+        buffer <<  announce_ok;
+
+        std::vector<uint8_t> net_data = buffer.front(buffer.size());
+
+        _logger->debug << "Sending ANNOUNCE OK to conn_id: " << conn_ctx.conn_id << std::flush;
+
+        send_ctrl_msg(conn_ctx, buffer.front(buffer.size()));
```
GhostofCookie, line 204: Same comment, copy paste error?

GhostofCookie: I see this a few more times.

src/moq_instance.cpp
```diff
@@ -0,0 +1,1687 @@
...
+                                       << ", setting send"
+                                          ""
+                                          " state to ready"
+                                       << std::flush;
+
+                        // Indicate send is ready upon subscribe
+                        // TODO(tievens): Maybe needs a delay as subscriber may have not received ok before data is sent
+                        ptd->lock()->setSubscribeId(msg.subscribe_id);
+                        ptd->lock()->setSendStatus(MoQTrackDelegate::TrackSendStatus::OK);
+                        ptd->lock()->cb_sendReady();
```
GhostofCookie, line 406: Why not retrieve the lock once and store temporarily?

cmd/moq-example/server.cpp
```diff
@@ -0,0 +1,523 @@
...
+        if (!is_new) {
+            _logger->info << "Received announce from conn_id: " << conn_id
+                          << " for namespace_hash: " << track_namespace_hash
+                          << " is duplicate, ignoring" << std::flush;
+            return true;
+        }
+
+        // true results in Send announce OK
+        return true;
+    }
```
GhostofCookie, line 229: May not matter, but this never returns false?

cmd/moq-example/client.cpp
```diff
@@ -0,0 +1,396 @@
...
+public:
+    clientDelegate(const cantina::LoggerPointer& logger) :
+        _logger(std::make_shared<cantina::Logger>("MID", logger)) {}
+
+    void cb_newConnection(qtransport::TransportConnId conn_id,
+                          const std::span<uint8_t>& endpoint_id,
+                          const qtransport::TransportRemote& remote) override {}
+
+    void cb_connectionStatus(qtransport::TransportConnId conn_id,
+                             const std::span<uint8_t>& endpoint_id,
```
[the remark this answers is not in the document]
TimEvens: Thanks, updated.

cmd/moq-example/client.cpp
TimEvens: Should be fine to do that. Just updated.

src/moq_instance.cpp
TimEvens: thanks for catching that. It was left over and should have been removed. I've removed it.

src/moq_instance.cpp
TimEvens: updated.

cmd/moq-example/client.cpp
```diff
@@ -0,0 +1,396 @@
...
+    uint64_t object_id { 0 };
+
+    while (not stop) {
+        if (!published_track && qclient_vars::conn_id) {
+            _logger->info << "Publish track: " << t_namespace << "/" << t_name << std::flush;
+            mi->publishTrack(*qclient_vars::conn_id, track_delegate);
+            published_track = true;
+        }
+
+        if (track_delegate->statusSend() != quicr::MoQTrackDelegate::TrackSendStatus::OK) {
```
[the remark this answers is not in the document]
TimEvens: Replaced it with `getSendStatus()`

src/moq_instance.cpp
TimEvens: Updated.

cmd/moq-example/server.cpp
TimEvens: Yeah, it doesn't matter right now.  It's for the future if we did some auth/validation, which would then return false.  Same thing for subscribe.

src/moq_instance.cpp
```diff
@@ -0,0 +1,1672 @@
...
+    MoQTrackDelegate::SendError MoQInstance::send_object(std::weak_ptr<MoQTrackDelegate> track_delegate,
+                                                         uint8_t priority,
+                                                         uint32_t ttl,
+                                                         bool stream_header_needed,
+                                                         uint64_t group_id,
+                                                         uint64_t object_id,
+                                                         std::span<const uint8_t> data)
+    {
+
+        auto td = track_delegate.lock();
```
GhostofCookie, line 1313: Should we check this isn't null, just in case?

GhostofCookie: To that end, should we just pass a shared_ptr? Or do we not want it to be kept alive by this method?

GhostofCookie: I think locking at the beginning and trusting it isn't nullptr is equivalent to just passing the shared_ptr by value. If we don't want the copy of the shared_ptr to share ownership, then maybe we should just pass a shared_ptr reference?

src/moq_instance.cpp
```diff
@@ -0,0 +1,1672 @@
...
+    }
+
+    void MoQInstance::on_recv_stream(const TransportConnId& conn_id,
+                        uint64_t stream_id,
+                        std::optional<DataContextId> data_ctx_id,
+                        const bool is_bidir)
+    {
+        auto stream_buf = _transport->getStreamBuffer(conn_id, stream_id);
+
+        //std::lock_guard<std::mutex> _(_state_mutex);
```
GhostofCookie, line 1525: Is this intentional?

src/moq_instance.cpp
```diff
@@ -0,0 +1,1672 @@
...
+            if (not data_ctx_id) {
+                close_connection(conn_id,
+                                 MoQTerminationReason::INTERNAL_ERROR,
+                                 "Received bidir is missing data context");
+                return;
+            }
+            conn_ctx.ctrl_data_ctx_id = data_ctx_id;
+        }
+
+        for (int i=0; i < 60; i++) { // don't loop forever, especially on bad stream
```
GhostofCookie, line 1543: Why 60?

src/moq_instance.cpp
```diff
@@ -0,0 +1,1672 @@
...
+            else {
+                if (process_recv_stream_data_message(conn_ctx, stream_buf)) {
+                    break;
+                }
+            }
+
+
+            if (!stream_buf->size()) { // done
+                break;
+            }
```
GhostofCookie, line 1574: I feel like some reordering of if statements and compressing of conditions might flatten the tabbing on this a bit... but the exact layout is escaping me, so this should be fine.

include/quicr/moq_instance_delegate.h
```diff
@@ -0,0 +1,133 @@
...
+        /**
+         * @brief Server Mode; Notification on new connection
+         * @details Notification that a new connection has been accepted
+         *
+         * @param conn_id          Transport connection ID
+         * @param endpoint_id      Endpoint ID of client connection
+         * @param remote           Transport remote connection information
+         */
+        virtual void cb_newConnection([[maybe_unused]] TransportConnId conn_id,
+                                      [[maybe_unused]] const std::span<uint8_t> endpoint_id,
```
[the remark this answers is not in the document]
TimEvens: Thanks, that should have been changed with the others. Updated now.

include/quicr/moq_track_delegate.h
```diff
@@ -0,0 +1,318 @@
...
+         * @brief set/update the track mode for sending
+         */
+        void setTrackMode(TrackMode track_mode) { _mi_track_mode = track_mode; }
+
+        // --------------------------------------------------------------------------
+        // Public Virtual API callback event methods to be overridden
+        // --------------------------------------------------------------------------
+
+        /**
+         * @brief Notificaiton of received data object
```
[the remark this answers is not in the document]
TimEvens: Updated.

src/moq_instance.cpp
```diff
@@ -0,0 +1,1672 @@
...
+                                  << " name_hash: " << th.track_name_hash
+                                  << std::flush;
+
+                    stream_buffer->resetAny();
+                    return true;
+                }
+
+                break;
+            }
+            case MoQMessageType::GOWAY: {
```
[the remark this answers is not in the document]
TimEvens: Good catch, that was a type in moq_message.h

src/moq_instance.cpp
TimEvens: Moq instance, at least right now, controls setting this from a shared pointer, which is guaranteed to not be null.  In this sense it's known and doesn't need to be checked if null, but I suppose it wouldn't impact it if we checked if it was null or not, but it would be error if it was nullptr, which probably should result in an abort.

src/moq_instance.cpp
TimEvens: Yes, locking is applied during the operation, so it was redundant.  I had it there initially and I want to go back to see if we should move the locking around a little bit.   

I'll update the comment.   Now that it's working, I would like to see if we can further to optimize the locking in a different PR.

src/moq_instance.cpp
TimEvens: It's just a number in the sand.  I'll add a const, but it'll still be some number greater than 50 and less than 200.  It's there to ensure that the thread doesn't starve out other streams, which does happen if you just use a while loop.  The reason it does need to be more than 50 is to support packet bursts.

TimEvens: Updated.

src/moq_instance.cpp
TimEvens: It is a bit funky due to the stream processing of not always being able to get the uintvar.  Only if we actually get it, will it be set. Otherwise, it'll remain unset.

src/moq_messages.cpp
```diff
@@ -0,0 +1,1345 @@
...
+      [[fallthrough]];
+    }
+    default:
+      break;
+  }
+
+  if(!msg.parse_completed) {
+    return false;
+  }
+  return true;
```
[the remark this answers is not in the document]
TimEvens: Updated

include/quicr/moq_messages.h
```diff
@@ -0,0 +1,409 @@
...
+  Path = 0x1,
+  AuthorizationInfo = 0x2, // version specific, unused
+  EndpointId = 0xF0, // Endpoint ID, using temp value for now
+  Invalid = 0xFF, // used internally.
+};
+
+struct MoqParameter {
+  uint64_t param_type {0};
+  uint64_t param_length {0};
+  bytes param_value;
```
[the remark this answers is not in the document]
TimEvens: Updated.

GhostofCookie (approved):

PR #180, opened 2024-08-14
Start on re-enabling clang-tidy & linting

RichLogan:
- Rename CLANG_TIDY option to LINT for consistency with other quicr libraries (plus we may add other non-clang-tidy linting steps).
- Disable checks to re-enable sequentially.
- Update transport for resolved checks.
- Enable LINT on CI.

This commit is expected to fail!

(Needs https://github.com/Quicr/transport/pull/168)


benchmark/CMakeLists.txt
```diff
@@ -29,7 +29,7 @@ set_target_properties(quicr_benchmark
         CXX_STANDARD_REQUIRED YES
         CXX_EXTENSIONS ON)
 
-if(CLANG_TIDY_EXE)
+if(LINT)
     set_target_properties(quicr_benchmark PROPERTIES CXX_CLANG_TIDY "${CLANG_TIDY_EXE}")
 endif()
```
GhostofCookie, line 34: Might be good to keep the original check as well, in case clang-tidy is not installed?

cmd/moq-example/CMakeLists.txt
```diff
@@ -21,7 +21,7 @@ set_target_properties(qserver
         CXX_STANDARD_REQUIRED YES
         CXX_EXTENSIONS ON)
 
-if(CLANG_TIDY_EXE)
+if(LINT)
   set_target_properties(quicr PROPERTIES CXX_CLANG_TIDY "${CLANG_TIDY_EXE}")
 endif()
```
GhostofCookie, line 26: Same as previous comment.

cmd/really/CMakeLists.txt
```diff
@@ -14,7 +14,7 @@ set_target_properties(really
         CXX_STANDARD_REQUIRED YES
         CXX_EXTENSIONS ON)
 
-if(CLANG_TIDY_EXE)
+if(LINT)
   set_target_properties(quicr PROPERTIES CXX_CLANG_TIDY "${CLANG_TIDY_EXE}")
 endif()
```
GhostofCookie, line 19: Ditto.

benchmark/CMakeLists.txt
RichLogan: Yeah I had debated this actually. I did change it to REQUIRED so in theory it MUST exist but think on review I'll actually add a cmake function to set this on our targets for consistency and can check it exists there.

RichLogan:
@GhostofCookie see https://github.com/Quicr/transport/pull/168 and the new changes for a new proposal.

PR #182, opened 2024-08-20
Initial MOQT based APIs

### LINKED ISSUE #178: MoQT API Cullen Comments
Here's a PDF version of Cullen's comments including response/questions/etc.

[cullens_questions_responses.pdf](https://github.com/user-attachments/files/16553930/cullens_questions_responses.pdf)

A list of agreed changes from Cullen's comments. The trailing (#) are the corresponding comment number.

- [x] Combine MoQClient and MoQClientDelegate into a single class. (1)
- [x] Remove cantina::Logger. (2)
- [x] metrics_export shoud not include 'moqt_config.h'. (5)
- [x] Client run method should be renamed - might be connect() and also have a correlating disconnect(). (6)
- [ ] Create outline of the FSM and make sure status of each state the client software needs to be aware of. (10)
- [x] Add a new method to send an announce (11) - Does not make sense for server, but for client.... keep in mind that the publishTrack handler has state and is updated by the implementation when there are subscribers or when subscribers leave. Callbacks are used to notify the app to start/stop their side, such as enable encoder or not. Why would a producer send an announce without knowing the track names that it will be producing? 
- [ ] #183
- [x] Pass access to client/server object as part of the callback for PublishTrackHandler and SubscribeTrackHandler (17) - No longer needed because client and server callbacks are within the client and server class now. 
- [x] Remove "setTrackMode()" (19)
- [x] Wrap the object meta-data and data in a class for sending/receiving. (20)
- [x] Rename TrackFullName to FullTrackName. Expand class to include Namespace, Name, possibly tuples in the future, etc. (21)
- [x] Rename TrackSubscribeHandler to SubscribeTrackHandler, same for publisher. (23)
- [x] Server and ServerCallbacks should be combined. (25) **DUP of (1)**
- [-] Callback in server callbacks seems not aligned with other ones. (28) - Callbacks are relative to client vs server. They will change as MOQT draft changes.
- [x] TrackFullName should be a class call FullTrackName. Element should be trackName (not name). With optionsl trackAlias. (32)
- [x] Remove MOQT from all class/variable names. Use namespace `moq` instead. Move include/source under `moq/` (24)

TimEvens:
Please only review `include/moqt/*` and `docs/*` files.  Please ignore the PR build errors. This change is only for API header changes and api documentation that is available via https://quicr.github.io/libquicr/moq-api-html

>[!NOTE]
> Implementation is not part of this review. This review is only for public API's (public and protected members), hence private members and dependent includes are not part of the public API. Implementation has not been updated or changed yet to implement the public API changes. We will have other PRs for implementation. Changes to private members should not effect or change any public API client/server implementation. 

resolves #178


include/moqt/client.h
```diff
@@ -0,0 +1,85 @@
...
+namespace moq::transport {
+    using namespace qtransport;
+
+    /**
+     * @brief MoQT Client
+     *
+     * @details MoQT Client is the handler of the MoQT QUIC transport IP connection.
+     */
+    class Client : public Transport
+    {
```
suhasHere, line 21: my preference would be to add an api to set the transport instead .. This would enable dependency injection flows and enables easy testing with fakes for things

include/moqt/client.h
```diff
@@ -0,0 +1,85 @@
...
+        /**
+         * @brief Starts a client connection via a transport thread (non-blocking)
+         *
+         * @details Makes a client connection session and runs in a newly created thread. All control and track
+         *   callbacks will be run based on events.
+         *
+         * @return Status indicating state or error. If successful, status will be
+         *    kClientConnecting.
+         */
+        Status Connect();
```
suhasHere, line 44: Since connect's logic happens on a separate transport thread, is Status() is synchronous ? Also kClientConnecting doesn't mean successful though, isn't it ?

include/moqt/client.h
```diff
@@ -0,0 +1,85 @@
...
+
+        /**
+         * @brief Disconnect the client connection gracefully (blocking)
+         *
+         * @details Unsubscribes and unpublishes all remaining active ones, sends MoQT control messages
+         *   for those and then closes the QUIC connection gracefully. Stops the transport thread. The class
+         *   destructor calls this method as well. Status will be updated to reflect not connected.
+         */
+        void Disconnect();
```
suhasHere, line 53: Is this a sync api ?

include/moqt/client.h
```diff
@@ -0,0 +1,85 @@
...
+
+        /**
+         * @brief Callback notification for connection status/state change
+         * @details Callback notification indicates state change of connection, such as disconnected
+         *
+         * @param status           Transport status of connection id
+         */
+        virtual void ConnectionChanged(TransportStatus status) = 0;
```
suhasHere, line 61: s/ConnectionChanged/OnConnectiontStatusChanged ??

include/moqt/client.h
```diff
@@ -0,0 +1,85 @@
...
+
+        /**
+         * @brief Callback on server setup message
+         * @details Server will send sever setup in response to client setup message sent. This callback is called
+         *  when a server setup has been received.
+         *
+         * @param server_setup     Decoded sever setup message
+         */
+        virtual void ServerSetup(const messages::MoqServerSetup& server_setup) = 0;
```
suhasHere, line 70: How is the client expected to use this callback ? A note on it might be helpful.

Also for callbacks can we do the form of "onXXXXYYYZZZ(..)"

include/moqt/client.h
```diff
@@ -0,0 +1,85 @@
...
+        /**
+         * @brief Notification callback to provide sampled metrics
+         *
+         * @details Callback will be triggered on Config::metrics_sample_ms to provide the sampled data based
+         *      on the sample period.  After this callback, the period/sample based metrics will reset and start over
+         *      for the new period.
+         *
+         * @param metrics           Copy of the connection metrics for the sample period
+         */
+        virtual void MetricsSampled(const ConnectionMetrics&& metrics)  = 0;
```
suhasHere, line 81: I thought we will go with a poll api for Metrics ? I think we might want to add a poll API which is async and gets resolved via this callback ?
 Also s/MetricsSampled/OnConnectionMetrics/

suhasHere: General Comment for the callbacks
- Let's add a note on the thread where this gets invoked
- Let's add a note to the API user that the control needs to relinquished to the thread asap .. (if that's the expectation)
- Change the names to "OnXXXYYY(....) form

include/moqt/common.h
```diff
@@ -0,0 +1,75 @@
...
+#pragma once
+
+#include <string>
+#include <transport/transport.h>
+
+namespace moq::transport {
+
+    constexpr uint64_t kMoqtVersion = 0xff000004; ///< draft-ietf-moq-transport-04
+    constexpr uint64_t kSubscribeExpires = 0;     ///< Never expires
+    constexpr int kReadLoopMaxPerStream = 60; ///< Support packet/frame bursts, but do not allow starving other streams
```
suhasHere, line 16: can we specify the units for 60 in the variable name

include/moqt/common.h
```diff
@@ -0,0 +1,75 @@
...
+
+    using namespace qtransport;
+
+    using Bytes = std::vector<uint8_t>;
+    using BytesSpan = Span<uint8_t const>;
```
suhasHere, line 21: My preference would be not use aliasing here .. as it seems confusing

include/moqt/common.h
```diff
@@ -0,0 +1,75 @@
...
+     *
+     * @details Struct of the full track name, which includes the namespace tuple, name, and track alias
+     *   Track alias will be set by the Transport.
+     */
+    struct FullTrackName
+    {
+        Span<uint8_t const> name_space;
+        Span<uint8_t const> name;
+        std::optional<uint64_t> track_alias;
+    };
```
suhasHere, line 34: let's make it a class and add functions . 
s/name_space/track_namespace. Would be easier of namespace was not spit.
s/name/track_name;
track_alias is not optional.

include/moqt/common.h
```diff
@@ -0,0 +1,75 @@
...
+
+    /**
+     * @brief Object headers struct
+     *
+     * @details Object headers are passed when sending and receiving an object. The object headers describe the object.
+     */
+    struct ObjectHeaders
+    {
+        uint64_t group_id;                ///< Object group ID - Should be derived using time in microseconds
```
suhasHere, line 68: let's remove the comment as it is not always time.

include/moqt/common.h
```diff
@@ -0,0 +1,75 @@
...
+        uint64_t object_id;               ///< Object ID - Start at zero and increment for each object in group
```
suhasHere, line 69: same here. I would remove the comment

include/moqt/common.h
```diff
@@ -0,0 +1,75 @@
...
+        uint64_t payload_length;          ///< Length of payload of the object data
+        std::optional<uint32_t> priority; ///< Priority of the object, lower value is better
```
suhasHere, line 71: priority is non-optional field.

include/moqt/core/base_track_handler.h
```diff
@@ -0,0 +1,138 @@
...
+
+        BaseTrackHandler() = delete;
+
+      protected:
+        /**
+         * @brief Track delegate constructor
+         *
+         * @param full_track_name       Full track name struct
+         */
+        BaseTrackHandler(const FullTrackName& full_track_name)
```
suhasHere, line 52: i think we need a constructor to take in tuples for namespace , byte string for name

include/moqt/core/base_track_handler.h
```diff
@@ -0,0 +1,138 @@
...
+        /**
+         * @brief Set the track alias
+         * @details MOQ transport instance will set the track alias when the track has
+         *   been assigned.
+         *
+         * @param track_alias       MoQT track alias for track namespace+name that
+         *                          is relative to the QUIC connection session
+         */
+        void SetTrackAlias(uint64_t track_alias) { full_track_name_.track_alias = track_alias; }
+
```
suhasHere, line 70: we should add a note that this api will override the existing value. Also is it OK to call this api. mid-sessopn ?

include/moqt/core/base_track_handler.h
```diff
@@ -0,0 +1,138 @@
...
+        /**
+         * @brief Get the track alias
+         * @returns Track alias as an optional. Track alias may not be set yet. If not
+         *   set, nullopt will be returned.
+         */
+        std::optional<uint64_t> GetTrackAlias() { return full_track_name_.track_alias; }
```
suhasHere, line 76: should be a const function

include/moqt/core/base_track_handler.h
```diff
@@ -0,0 +1,138 @@
...
+
+        /**
+         * @brief Sets the subscribe ID
+         * @details MoQ instance sets the subscribe id based on subscribe track method call. Subscribe
+         *      id is specific to the connection, so it must be set by the moq instance/connection.
+         *
+         * @param subscribe_id          62bit subscribe ID
+         */
+        void GetSubscribeId(std::optional<uint64_t> subscribe_id) { subscribe_id_ = subscribe_id; }
```
suhasHere, line 85: Shouldn't this be SetSubcribeId() ?

include/moqt/core/base_track_handler.h
```diff
@@ -0,0 +1,138 @@
...
+
+        /**
+         * @brief Get the subscribe ID
+         *
+         * @return nullopt if not subscribed, otherwise the subscribe ID
+         */
+        std::optional<uint64_t> GetSubscribeId() { return subscribe_id_; }
```
suhasHere, line 92: const function

include/moqt/core/base_track_handler.h
```diff
@@ -0,0 +1,138 @@
...
+
+        /**
+         * @brief Get the full track name
+         *
+         * @details Gets the full track name
+         *
+         * @return FullTrackName
+         */
+        FullTrackName GetFullTrackName() { return { full_track_name_ }; }
```
suhasHere, line 101: return full_track_name

include/moqt/core/base_track_handler.h
```diff
@@ -0,0 +1,138 @@
...
+
+        /**
+         * @brief Get the connection ID
+         */
+        uint64_t GetConnectionId() { return conn_id_; };
```
suhasHere, line 106: const function

include/moqt/core/base_track_handler.h
```diff
@@ -0,0 +1,138 @@
...
+         * @details The MOQ Handler sets the connection ID
+         */
+        void SetConnectionId(uint64_t conn_id) { conn_id_ = conn_id; };
+
+        // --------------------------------------------------------------------------
+        // Member variables
+        // --------------------------------------------------------------------------
+
+        const FullTrackName full_track_name_;
+        uint64_t conn_id_;
```
suhasHere, line 124: conn_id_{0}

include/moqt/core/base_track_handler.h
```diff
@@ -0,0 +1,138 @@
...
+        /**
+         * subscribe_id_ is the primary index/key for subscribe subscribe context/delegate storage.
+         *   It is use as the subscribe_id in MoQT related subscribes.  Subscribe ID will adapt
+         *   to received subscribe IDs, so the value will reflect either the received subscribe ID
+         *   or the next one that increments from last received ID.
+         */
+        std::optional<uint64_t> subscribe_id_;
+
+        uint64_t prev_object_group_id_{ 0 };
+        uint64_t prev_object_id_{ 0 };
```
suhasHere, line 135: need comments on these to help understand their role

include/moqt/core/transport.h
```diff
@@ -0,0 +1,226 @@
...
+     */
+    class Transport : public ITransport::TransportDelegate
+    {
+      public:
+        Transport() = delete;
+
+        enum class Status : uint8_t
+        {
+            kReady = 0,
+            kNotReady,
```
suhasHere, line 39: what causes the status to go kNotReady ?

include/moqt/core/transport.h
```diff
@@ -0,0 +1,226 @@
...
+            kClientNotConnected,
+            kClientFailedToConnect
+        };
+
+        /**
+         * @brief Client mode Constructor to create the MOQ instance
+         *
+         * @param cfg       MoQT Instance Client Configuration
+         */
+        Transport(const ClientConfig& cfg);
```
suhasHere, line 55: can we mark it explcit

include/moqt/core/transport.h
```diff
@@ -0,0 +1,226 @@
...
+        // -------------------------------------------------------------------------------------------------
+        /**
+         * @brief Subscribe to a track
+         *
+         * @param conn_id           Connection ID to send subscribe
+         * @param track_delegate    Track delegate to use for track related functions and callbacks
+         *
+         * @returns `track_alias` if no error and nullopt on error
+         */
+        std::optional<uint64_t> SubscribeTrack(TransportConnId conn_id,
```
suhasHere, line 77: this api seems to be in wrong layer/class

include/moqt/core/transport.h
```diff
@@ -0,0 +1,226 @@
...
+                                               std::shared_ptr<SubscribeTrackHandler> track_delegate);
+
+        /**
+         * @brief Unsubscribe track
+         *
+         * @param conn_id           Connection ID to send subscribe
+         * @param track_delegate    Track delegate to use for track related functions and callbacks
+         */
+        void UnsubscribeTrack(TransportConnId conn_id, std::shared_ptr<SubscribeTrackHandler> track_delegate);
+
```
suhasHere, line 87: same comment as above

include/moqt/core/transport.h
```diff
@@ -0,0 +1,226 @@
...
+         *
+         * @param conn_id           Connection ID from transport for the QUIC connection context
+         * @param track_delegate    Track delegate to use for track related functions
+         *                          and callbacks
+         *
+         * @returns `track_alias` if no error and nullopt on error
+         */
+        std::optional<uint64_t> PublishTrack(TransportConnId conn_id,
+                                             std::shared_ptr<PublishTrackHandler> track_delegate);
+
```
suhasHere, line 99: similar comment as above.
Publish and Subscribe should not be on transport .

include/moqt/core/transport.h
```diff
@@ -0,0 +1,226 @@
...
+        {
+        }
+
+        void on_connection_status(const TransportConnId& conn_id, const TransportStatus status) override;
+        void on_new_connection(const TransportConnId& conn_id, const TransportRemote& remote) override;
+        void on_recv_stream(const TransportConnId& conn_id,
+                            uint64_t stream_id,
+                            std::optional<DataContextId> data_ctx_id,
+                            const bool is_bidir = false) override;
+        void on_recv_dgram(const TransportConnId& conn_id, std::optional<DataContextId> data_ctx_id) override;
```
suhasHere, line 143: nit s/dgram/datagram

include/moqt/core/transport.h
```diff
@@ -0,0 +1,226 @@
...
+                                 [[maybe_unused]] const DataContextId& data_ctx_id) override
+        {
+        }
+
+        void on_connection_status(const TransportConnId& conn_id, const TransportStatus status) override;
+        void on_new_connection(const TransportConnId& conn_id, const TransportRemote& remote) override;
+        void on_recv_stream(const TransportConnId& conn_id,
+                            uint64_t stream_id,
+                            std::optional<DataContextId> data_ctx_id,
+                            const bool is_bidir = false) override;
```
suhasHere, line 142: callback typically shouldn't have default parameters.

include/moqt/core/transport.h
```diff
@@ -0,0 +1,226 @@
...
+         * @param track_delegate    Track delegate used when published track
+         */
+        void UnpublishTrack(TransportConnId conn_id, std::shared_ptr<PublishTrackHandler> track_delegate);
+
+        /**
+         * @brief Get the instance status
+         *
+         * @return Status indicating the state/status of the instance
+         */
+        Status GetStatus();
```
suhasHere, line 113: There is Status and TransportStatus within Transport. Wondering if we need to merge them or rename it to better reflect the context

include/moqt/core/transport.h
```diff
@@ -0,0 +1,226 @@
...
+        // -------------------------------------------------------------------------------------------------
+
+        struct ConnectionContext
+        {
+            TransportConnId conn_id;
+            std::optional<uint64_t> ctrl_data_ctx_id;
+            bool setup_complete{ false }; ///< True if both client and server setup messages have completed
+            uint64_t client_version{ 0 };
+            std::optional<messages::MoqtMessageType>
+              ctrl_msg_type_received; ///< Indicates the current message type being read
```
suhasHere, line 156: can it be data message too ? If so we should rename the variable name

include/moqt/core/transport.h
```diff
@@ -0,0 +1,226 @@
...
+
+            uint64_t current_subscribe_id{ 0 }; ///< Connection specific ID for subscribe messages
+
+            /// Track namespace/name by received subscribe IDs
+            /// Used to map published tracks to subscribes in client mode
+            std::map<uint64_t, std::pair<uint64_t, uint64_t>> recv_sub_id;
```
suhasHere, line 162: should this be a map of subscribe_id to FullTrackName here ?

include/moqt/core/transport.h
```diff
@@ -0,0 +1,226 @@
...
+
+            /// Tracks by subscribe ID
+            std::map<uint64_t, std::shared_ptr<BaseTrackHandler>> tracks_by_sub_id;
+
+            /// Publish tracks by namespace and name. map[track namespace][track name] = track delegate
+            std::map<uint64_t, std::map<uint64_t, std::shared_ptr<BaseTrackHandler>>> pub_tracks_by_name;
```
suhasHere, line 168: should we use FullTrackName instead ?

include/moqt/core/transport.h
```diff
@@ -0,0 +1,226 @@
...
+#include <string_view>
+
+namespace moq::transport {
+
+    /**
+     * @brief MOQ Implementation supporting both client and server modes
+     * @details MoQ implementation is the handler for either a client or server. It can run
+     *   in only one mode, client or server.
+     */
+    class Transport : public ITransport::TransportDelegate
```
suhasHere, line 31: wonder if the name of this class being something like MoQSession ?

include/moqt/core/transport.h
```diff
@@ -0,0 +1,226 @@
...
+        void on_recv_dgram(const TransportConnId& conn_id, std::optional<DataContextId> data_ctx_id) override;
+
+        // -------------------------------------------------------------------------------------------------
+        // End of transport delegate/callback functions
+        // -------------------------------------------------------------------------------------------------
+
+        struct ConnectionContext
+        {
+            TransportConnId conn_id;
+            std::optional<uint64_t> ctrl_data_ctx_id;
```
suhasHere, line 152: why is it optional ? It may help identify what is a data context ?

include/moqt/core/transport.h
```diff
@@ -0,0 +1,226 @@
...
+                             const std::string& reason_str);
+        bool ProcessRecvCtrlMessage(ConnectionContext& conn_ctx, std::shared_ptr<StreamBuffer<uint8_t>>& stream_buffer);
+        bool ProcessRecvStreamDataMessage(ConnectionContext& conn_ctx,
+                                          std::shared_ptr<StreamBuffer<uint8_t>>& stream_buffer);
+
+        void RemoveSubscribeTrack(ConnectionContext& conn_ctx,
+                                  SubscribeTrackHandler& delegate,
+                                  bool remove_delegate = true);
+
+        std::optional<std::weak_ptr<PublishTrackHandler>> GetPubTrackHandler(ConnectionContext& conn_ctx,
```
suhasHere, line 202: given it is a weak pointer, i don;t see it needs to be inside an optional.

include/moqt/core/transport.h
```diff
@@ -0,0 +1,226 @@
...
+            std::map<uint64_t, std::shared_ptr<BaseTrackHandler>> tracks_by_sub_id;
+
+            /// Publish tracks by namespace and name. map[track namespace][track name] = track delegate
+            std::map<uint64_t, std::map<uint64_t, std::shared_ptr<BaseTrackHandler>>> pub_tracks_by_name;
+        };
+
+        // -------------------------------------------------------------------------------------------------
+        // Private methods
+        // -------------------------------------------------------------------------------------------------
+        void Init();
```
suhasHere, line 174: comment on what does Init() do might be helpful

include/moqt/core/transport.h
```diff
@@ -0,0 +1,226 @@
...
+
+        // -------------------------------------------------------------------------------------------------
+        // Private member variables
+        // -------------------------------------------------------------------------------------------------
+
+        std::mutex state_mutex_;
+        const bool client_mode_;
+        bool stop_{ false };
+        const ServerConfig server_config_;
+        const ClientConfig client_config_;
```
suhasHere, line 213: these needs to be optional as it is set only in one of the modes. May be I don't understand the need for it to be constants

include/moqt/core/transport.h
```diff
@@ -0,0 +1,226 @@
...
+
+        std::map<TransportConnId, ConnectionContext> connections_;
+
+        Status status_{ Status::kNotReady };
+
+#ifndef LIBQUICR_WITHOUT_INFLUXDB
+        quicr::MetricsExporter mexport_;
+#endif
```
suhasHere, line 221: let's move exporters out of the library

include/moqt/publish_track_handler.h
```diff
@@ -0,0 +1,267 @@
...
+
+            /// Previous object payload has not been completed and new object cannot start in per-track track mode
+            /// without creating a new track. This requires to unpublish and to publish track again.
+            kPreviousObjectNotCompleteMustStartNewTrack,
+        };
+
+        /**
+         * @brief  Status codes for the publish track
+         */
+        enum class Status : uint8_t
```
suhasHere, line 61: Should we call it TrackStatus ?

include/moqt/publish_track_handler.h
```diff
@@ -0,0 +1,267 @@
...
+     */
+    class PublishTrackHandler : protected BaseTrackHandler
+    {
+      public:
+        friend class Transport;
+
+        /**
+         * @brief Publish status codes
+         */
+        enum class PublishStatus : uint8_t
```
suhasHere, line 30: rename it to PublishResult ?

include/moqt/publish_track_handler.h
```diff
@@ -0,0 +1,267 @@
...
+        {
+            kOk = 0,
+            kInternalError,
+            kNotAuthorized,
+            kNotAnnounced,
```
suhasHere, line 35: above 2 might be better scoped to Track status

include/moqt/publish_track_handler.h
```diff
@@ -0,0 +1,267 @@
...
+            /// any/all data that was sent already.
+            kObjectDataTooLarge,
+
+            /// Previous object payload has not been completed and new object cannot start in per-group track mode
+            /// unless new group is used.
+            kPreviousObjectNotCompleteMustStartNewGroup,
+
+            /// Previous object payload has not been completed and new object cannot start in per-track track mode
+            /// without creating a new track. This requires to unpublish and to publish track again.
+            kPreviousObjectNotCompleteMustStartNewTrack,
```
suhasHere, line 55: I don't understand the last 2 errors

include/moqt/publish_track_handler.h
```diff
@@ -0,0 +1,267 @@
...
+            kPendingAnnounceResponse,
+            kAnnounceNotAuthorized,
+            kNoSubscribers,
+            kSendingUnannounce,
+        };
+
+
+        // --------------------------------------------------------------------------
+        // Public API methods that normally should not be overridden
+        // --------------------------------------------------------------------------
```
suhasHere, line 75: Curios, Why is the restriction to not overridde and what issues happen if one does override ?

include/moqt/publish_track_handler.h
```diff
@@ -0,0 +1,267 @@
...
+         *
+         * @param full_track_name       Full track name struct
+         * @param track_mode            The track mode to operate using
+         * @param default_priority      Default priority for objects if not specified in ObjectHeaderss
+         * @param default_ttl           Default TTL for objects if not specified in ObjectHeaderss
+         */
+        PublishTrackHandler(const FullTrackName& full_track_name,
+                            TrackMode track_mode,
+                            uint8_t default_priority,
+                            uint32_t default_ttl)
```
suhasHere, line 88: may be remove the default constructor

include/moqt/publish_track_handler.h
```diff
@@ -0,0 +1,267 @@
...
+          : BaseTrackHandler(full_track_name)
+          , track_mode_(track_mode)
+          , def_priority_(default_priority)
```
suhasHere, line 91: s/def/default

include/moqt/publish_track_handler.h
```diff
@@ -0,0 +1,267 @@
...
+          , def_ttl_(default_ttl)
```
suhasHere, line 92: s/def/default

include/moqt/publish_track_handler.h
```diff
@@ -0,0 +1,267 @@
...
+        {
+        }
+
+        // --------------------------------------------------------------------------
+        // Public Virtual API callback event methods to be overridden
+        // --------------------------------------------------------------------------
+
+        /**
+         * @brief Notification of publish status
```
suhasHere, line 101: this seems to be not PublishStatus field but Status field in the base class.

include/moqt/publish_track_handler.h
```diff
@@ -0,0 +1,267 @@
...
+         * @details Notification of publishing status, such as when it's ready to publish or not ready to publish
+         *
+         * @param status        Indicates the status of being able to publish
+         */
+        virtual void StatusChanged(PublishTrackHandler::Status status) = 0;
```
suhasHere, line 106: rename to be called OnTrackStatusChanged ??

include/moqt/publish_track_handler.h
```diff
@@ -0,0 +1,267 @@
...
+        // --------------------------------------------------------------------------
+        /**
+         * @brief set/update the default priority for published objects
+         */
+        void SetDefaultPriority(uint8_t priority) { def_priority_ = priority; }
+
+        /**
+         * @brief set/update the default TTL expiry for published objects
+         */
+        void SetDefaultTTL(uint32_t ttl) { def_ttl_ = ttl; }
```
suhasHere, line 131: const uiint32_t ttl

include/moqt/publish_track_handler.h
```diff
@@ -0,0 +1,267 @@
...
+         *
+         * @param object_headers        Object headers, must include group and object Ids
+         * @param data                  Payload data for the object, must be <= object_headers.payload_length
+         *
+         * @returns Publish status of the publish
+         *    * PublishStatus::kObjectContinuationDataNeeded if the object payload data is not completed but it was sent,
+         *    * PublishStatus::kObjectDataComplete if the object data was sent and the data is complete,
+         *    * other PublishStatus
+         */
+        PublishStatus PublishObject(const ObjectHeaders& object_headers, BytesSpan data);
```
suhasHere, line 175: not sure of the reason to use Span here. Isn't the entire object gets moved/copied over within the function ? Span is helpful when you want to read into and not for cases when you take over the ownership. May be I am missing how are the parameters used.
Also rename data to object_payload.

include/moqt/publish_track_handler.h
```diff
@@ -0,0 +1,267 @@
...
+         *                              the original object payload size.
+         *
+         * @returns
+         *      * PublishStatus::kObjectDataComplete if the data fulfills the remaining data,
+         *      * PublishStatus::kObjectContinuationDataNeeded if more data is still needed,
+         *      * PublishStatus::kObjectDataTooLarge if the data provided would exceed the remaining amount of data,
+         *      * PublishStatus::kNoPreviousObject if the previous object was completed or if there was no previous object,
+         *      * other PublishStatus
+         */
+        PublishStatus PublishContinuationData(BytesSpan data);
```
suhasHere, line 193: if the idea is to provide a streaming API , may be something like the below is less confusing for the application developer

 void publish(
      const ObjectHeaders& objHeader,
      uint64_t payloadOffset,
      std::vector<uint8_t>&& payload,
      bool end_of_message);

include/moqt/publish_track_handler.h
```diff
@@ -0,0 +1,267 @@
...
+        // --------------------------------------------------------------------------
+        Status publish_status_{ Status::kNotAnnounced };
+        TrackMode track_mode_;
+        uint8_t def_priority_;              // Set by caller and is used when priority is not specified
+        uint32_t def_ttl_;                  // Set by caller and is used when TTL is not specified
+
+        uint64_t publish_data_ctx_id_;      // publishing data context ID
+        PublishObjFunction publish_object_func_;
+
+        uint64_t object_payload_remaining_length_ { 0 };
```
suhasHere, line 263: which object are we referring to here. Personally think there is a lot more internal structure being exposed in the public header/API. These will be better suited in a implementation class ?

include/moqt/server.h
```diff
@@ -0,0 +1,144 @@
...
+
+namespace moq::transport {
+    using namespace qtransport;
+
+    /**
+     * @brief MoQT Server
+     *
+     * @details MoQT Server is the handler of the MoQT QUIC listening socket
+     */
+    class Server : public Transport
```
suhasHere, line 21: same comment as is Server a Transport or Does Server has a Transport handle

include/moqt/subscribe_track_handler.h
```diff
@@ -0,0 +1,156 @@
...
+         */
+        enum class Status : uint8_t
+        {
+            kOk = 0,
+            kNotConnected,
+            kSubscribeError,
+            kNotAuthorized,
+            kNotSubscribed,
+            kPendingSubscribeResponse,
+            kSendingUnsubscribe // Triggers callbacks to not be called in this state
```
suhasHere, line 49: There is some ambiguite between error and status. May be i don't fully grasp the intent and I might be wrong

suhasHere (commented):
Thanks for the PR @TimEvens  . I do a have nits ( feel free it ignore) and few comments on the API in general. I wasn't able to go through all the include flles. I hope my feedback will be useful.

include/moqt/core/transport.h
TimEvens: This is the MOQ Transport, which is integrates the underlining QUIC transport. From an API standpoint, this API does not directly interact with the QUIC transport, all that is provided within MoQT API.  A topic for discussion next week is around moving qtransport into libquicr.

include/moqt/core/transport.h
TimEvens: Private members, such as the above are not part of the public API.  These are internal use member variables that are for the implementation.

include/moqt/core/transport.h
TimEvens: Again, this is not part of the public API as these are private members used by the implementation.

include/moqt/core/transport.h
TimEvens: Again, this is not part of the public API as these are private members used by the implementation.

include/moqt/core/transport.h
TimEvens: Again, this is not part of the public API as these are private members used by the implementation.

include/moqt/core/transport.h
TimEvens: Again, this is not part of the public API as these are private members used by the implementation.

include/moqt/core/transport.h
TimEvens: Again, this is not part of the public API as these are private members used by the implementation.

include/moqt/core/transport.h
TimEvens: Again, this is not part of the public API as these are private members used by the implementation.

include/moqt/core/transport.h
TimEvens: Again, this is not part of the public API as these are private members used by the implementation.


But this is being removed, so moot.

include/moqt/publish_track_handler.h
TimEvens: Isn't that implied as this is status within `PublishTrackHandler()`?

include/moqt/publish_track_handler.h
TimEvens: Makes sense.

include/moqt/publish_track_handler.h
TimEvens: There is overlap on error/status conditions depending on which method you call. In this case, `PublishObjectStatus` is returned, which due to the async nature, you may no longer be authorized, etc.

include/moqt/publish_track_handler.h
TimEvens: So the descriptive comment to describe it does not do that? How would you suggest rewording it?

include/moqt/publish_track_handler.h
TimEvens: It doesn't matter. I'll remove that non public comment.

include/moqt/publish_track_handler.h
TimEvens: Implementation is not part of this review. This review is only for public API's. Implementation has not been updated or changed yet to implement the public API changes. Private members, such as the above are not part of the public API. These are internal use member variables that are for the implementation.

Will rename, but according to google style this is allowed in this scope.

include/moqt/publish_track_handler.h
TimEvens: It's for the publish track status, which is not the same as `PublishObjectStatus`.   

Updated the verbiage to: 

```c++
        /**
         * @brief Notification of publish track status change
         * @details Notification of a change to publish track status, such as
         *      when it's ready to publish or not ready to publish
         *
         * @param status        Indicates the status of being able to publish
         */
```

include/moqt/publish_track_handler.h
TimEvens: Updated.

include/moqt/publish_track_handler.h
TimEvens: No.  It's not copied/moved because that introduced unnecessary copies of the same object for each subscriber and  client. The construct of using span vs vector work transparently. You can pass a vector or you can pass `Bytes` another span `BytesSpan`. 

The implementation is not under review here, but the implementation handles this to ensure that when this method returns, the use of the object data is done.  We pass this down a few times to get it to the transport for transmission, that is why we had unnecessary copies, including the initial for each publish object fan out for all subscribers.  It is in the final transmission queue where we do a single copy for transmission. That isn't easily avoidable.

include/moqt/publish_track_handler.h
TimEvens: Why would you provide object headers repeatedly for continuation data?  End is also known by end of payload length requirement per https://github.com/moq-wg/moq-transport/blob/main/draft-ietf-moq-transport.md#object-stream.  

A topic for next week is that MoQT doesn't make it easy for object pipelining of unknown length data.  MoQT requires (https://github.com/moq-wg/moq-transport/blob/main/draft-ietf-moq-transport.md#object-stream) `payload_length` before you can even start the transmission of data bytes.  In draft-04 and draft-05 the payload length was not required for datagram or per-object streams, but that has changed in the recent updates (e.g., draft-06) that now requires it for all object messages, even datagram.  This basically results in the API having to require payload length before you can start to transmit.   Hence, we know the end based on the payload length.

include/moqt/server.h
TimEvens: `moq::transport::Transport` is common and used by both `moq::transport::Client` and `moq::transport::Server`. We could add `Handler` as a suffix to both `Client` and `Server` but that didn't seem to be needed as it gets weird with everything having handler in the name.

include/moqt/subscribe_track_handler.h
TimEvens: The comment actually should have been a proper doxygen comment, which it was missing the `/<`. I have updated both the publish and subscribe track handlers with the below for their relative sending unX states. 

```
///< In this state, callbacks will not be called
```

There is no error anymore, just status now.  The comment indicates that callbacks will not be called during this state. @shenning00  and I discussed this at length on handling tear down asynchronously gracefully for the application. When the client or server app does an `UnsubscribeTrack(...)` the client/server app would not want to receive callbacks anymore for the track that it just Unpublished or Unsubscribed.  This state indicates that unsubscribe (or also in unpublish) is in process and that callbacks will not be called... because that would be just weird to receive callbacks to a track that you are done with.  

We should talk more about this next week.

TimEvens:
@fluffy , @mzanaty , @shenning00, 


Please provide your review and comments to the discussions.  

>[!IMPORTANT]
> Implementation is not part of this review. This review is only for public API's (public and protected members), hence private members and dependent includes are not part of the public API. Implementation has not been updated or changed yet to implement the public API changes. We will have other PRs for implementation. Changes to private members should not effect or change any public API client/server implementation.

include/moqt/core/transport.h
suhasHere: I do understand this is not part of the api code to review but I did review since it was part of include/ which i feel still as public. Regardless, i was curious to understand if the comment makes sense in general ..

include/moqt/core/transport.h
suhasHere: The point here , this is not a Transport either. If we merge actual QUIC transport in this layer, there will be 2 different transport classes. May be Session is not the best name, but Transport is ambigouos too

include/moqt/publish_track_handler.h
suhasHere: It is more of makign it more readable

include/moqt/publish_track_handler.h
suhasHere: I don't see adding On makes it redundant. I still feel callback having On... is more readable

include/moqt/publish_track_handler.h
suhasHere: I was not advocating for copy but rather move .. Move semantics are not expensive. If we decide to not do that then the API should document saying the caller shouldn't free the resource until the function returns.

include/moqt/server.h
suhasHere: I see inheritance here, i was suggesting for composition

include/moqt/publish_track_handler.h
RichLogan: My comment is specifically about prefixing callbacks with the word `On`. My C# experience is leaking into this (what I'm about to describe is the C# event naming convention), but what I have always been used to is that the callback is named with a corresponding verb (with the timeline it occurs reflected in the tense) and that a function prefixed "On" (e.g `On<EventName>()`) normally *triggers* the corresponding callback, not the callback itself.

E.g example usage of both being:

```cpp
// Handler.
virtual void StatusChanged(PublishTrackHandler::Status status) = 0;

// Example Trigger (if exposed, not saying to add this).
void OnStatusChanged(PublishTrackHandler::Status status) {
    StatusChanged(status);
};
```

So I would vote leave it as it is.

PR #216, opened 2024-09-03
Option to build quicr as SHARED

RichLogan:


CMakeLists.txt
```diff
@@ -18,7 +18,7 @@ project(quicr
         LANGUAGES CXX)
 
 option(LINT "Perform linting with clang-tidy" OFF)
-
+option(QUICR_BUILD_SHARED "Build quicr as a SHARED library" OFF)
```
GhostofCookie, line 21: I think general cmake convention is to use lowercase names:
```suggestion
option(quicr_BUILD_SHARED "Build quicr as a SHARED library" OFF)
```

GhostofCookie (approved):

CMakeLists.txt
RichLogan: In this case I copied the format of the other options above, but maybe we can go change all of them.

PR #217, opened 2024-09-03
Cleanup & Refactor

GhostofCookie:
Cleanup and adding some helper functions, addressing some clang-tidy comments.

### DIFF
--- a/include/moq/detail/transport.h
+++ b/include/moq/detail/transport.h
@@ -116,7 +116,7 @@ namespace moq {
          * @param connection_handle         Connection ID to send subscribe
          * @param track_handler             Track handler to use for track related functions and callbacks
          */
-        void UnsubscribeTrack(ConnectionHandle connection_handle, std::shared_ptr<SubscribeTrackHandler> track_handler);
+        void UnsubscribeTrack(ConnectionHandle connection_handle, const std::shared_ptr<SubscribeTrackHandler>& track_handler);
 
         /**
          * @brief Publish to a track
@@ -133,7 +133,7 @@ namespace moq {
          * @param connection_handle           Connection ID from transport for the QUIC connection context
          * @param track_handler    Track handler used when published track
          */
-        void UnpublishTrack(ConnectionHandle connection_handle, std::shared_ptr<PublishTrackHandler> track_handler);
+        void UnpublishTrack(ConnectionHandle connection_handle, const std::shared_ptr<PublishTrackHandler>& track_handler);
 
         /**
          * @brief Bind a server publish track handler based on a subscribe
@@ -147,7 +147,7 @@ namespace moq {
          */
         void BindPublisherTrack(ConnectionHandle connection_handle,
                                 uint64_t subscribe_id,
-                                std::shared_ptr<PublishTrackHandler> track_handler);
+                                const std::shared_ptr<PublishTrackHandler>& track_handler);
 
         /**
          * @brief Get the status of the Client
@@ -228,7 +228,7 @@ namespace moq {
 
         void Init();
 
-        PublishTrackHandler::PublishObjectStatus SendObject(std::weak_ptr<PublishTrackHandler> track_handler,
+        PublishTrackHandler::PublishObjectStatus SendObject(const std::shared_ptr<PublishTrackHandler>& track_handler,
                                                             uint8_t priority,
                                                             uint32_t ttl,
                                                             bool stream_header_needed,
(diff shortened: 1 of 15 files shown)


include/moq/detail/transport.h
```diff
@@ -116,7 +116,7 @@ namespace moq {
          * @param connection_handle         Connection ID to send subscribe
          * @param track_handler             Track handler to use for track related functions and callbacks
          */
-        void UnsubscribeTrack(ConnectionHandle connection_handle, std::shared_ptr<SubscribeTrackHandler> track_handler);
+        void UnsubscribeTrack(ConnectionHandle connection_handle, const std::shared_ptr<SubscribeTrackHandler>& track_handler);
```
GhostofCookie, line 119: These were recommended by clang-tidy since we never use the copy of the shared ptrs. If this is the wrong assumption then let's talk about it.

src/CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 # SPDX-FileCopyrightText: Copyright (c) 2024 Cisco Systems
 # SPDX-License-Identifier: BSD-2-Clause
 
-add_library(quicr)
+add_library(quicr SHARED)
```
TimEvens, line 4: Moving to shared over static?

src/moq/transport.cpp
```diff
@@ -589,39 +567,39 @@ namespace moq {
            Span<const uint8_t> data) -> PublishTrackHandler::PublishObjectStatus {
            return SendObject(track_handler, priority, ttl, stream_header_needed, group_id, object_id, data);
        };
+       // Set the track handler for pub/sub
+
+       conn_it->second.pub_tracks_by_name[th.track_namespace_hash][th.track_name_hash] = std::move(track_handler);
    }
 
-   PublishTrackHandler::PublishObjectStatus Transport::SendObject(std::weak_ptr<PublishTrackHandler> track_handler,
+   PublishTrackHandler::PublishObjectStatus Transport::SendObject(const std::shared_ptr<PublishTrackHandler>& track_handler,
```
TimEvens, line 575: It's 100% safe using weak as this is self. Why make it shared?

src/moq/transport.cpp
```diff
@@ -1000,191 +978,146 @@ namespace moq {
...
+                                                msg.object_id,
+                                                msg.payload.size(),
+                                                msg.priority,
+                                                std::nullopt,
+                                                TrackMode::kStreamPerObject,
+                                                std::nullopt },
+                                              msg.payload);
+               stream_buffer->ResetAny();
+               return true;
                break;
```
TimEvens, line 1016: probably not needed considering return.

include/moq/detail/transport.h
TimEvens: It was moved before, but this works as well. We should be consistent here because the school of thought was that we should pass as a copy/ref count and then move within the method if that is the intent.

src/CMakeLists.txt
GhostofCookie: Oops, this conflicts with Rich's PR to make this optional for building in Decimus

src/moq/transport.cpp
GhostofCookie: Another recommendation of clang-tidy to use const ref weak_ptr, but that felt weird. The reality is that because we're using it so confidently as a weak ptr that we know will be valid, then we may as well pass the shared ptr since we are going to use it that way (so we save on `weak_ptr.lock()`), and since we don't store it we can pass it by const ref.

src/moq/transport.cpp
GhostofCookie: Oops!

include/moq/detail/transport.h
GhostofCookie: Yeah, but since we never moved then there's no point in copying the pointer, so pass by ref makes it clearer that we are not necessarily copying it (which we aren't).

src/moq/transport.cpp
RichLogan: If this function is always going to be called from a scope that ensures lifetime, what's wrong with a straight up const reference?

include/moq/detail/transport.h
RichLogan: This is another situation where this function doesn't involve lifetime, could it not just be a const ref?

src/moq/transport.cpp
```diff
@@ -587,41 +565,41 @@ namespace moq {
            uint64_t group_id,
            uint64_t object_id,
            Span<const uint8_t> data) -> PublishTrackHandler::PublishObjectStatus {
-           return SendObject(track_handler, priority, ttl, stream_header_needed, group_id, object_id, data);
+           return SendObject(*track_handler, priority, ttl, stream_header_needed, group_id, object_id, data);
        };
+       // Set the track handler for pub/sub
+
+       conn_it->second.pub_tracks_by_name[th.track_namespace_hash][th.track_name_hash] = std::move(track_handler);
```
TimEvens, line 572: Since track handler is passed by ref now, do we want `std::move()`?

include/moq/detail/transport.h
TimEvens: I suppose the `Un*` methods don't need shared pointer, but if we update one they all should be updated to be consistent.

src/moq/transport.cpp
TimEvens: Ignore. wrong method that I was referring to.

TimEvens (approved):
LGTM. We can change the Unpublish and Unsubscribe track methods in this PR or in another.

include/moq/detail/transport.h
GhostofCookie: Very possibly it could just be a const&. I'm not sure if changing it from shared_ptr makes it inconsistent with the PublishTrack and SubscribeTrack, which both use shared_ptrs (which there is no moving away from). Is it worth needing to dereference our shared_ptrs to use these Un* methods?

PR #227, opened 2024-09-09
Fix spdlog teardown issue

RichLogan:
Looks like the spdloggers created by `Transport` need to be explicitly torn down. Added a test that will demonstrate the issue when the `Transport` destructor is defaulted, and passes now that it drops the created logger.


GhostofCookie (approved):
This will fix Decimus for now, but we should inject the logger in another PR.

TimEvens (approved):
LGTM

PR #234, opened 2024-09-10
Version and supported archs

TimEvens:
Adds initial starting version that is defined in CMakeLists.txt.  `cmake` will generate `<bin dir>/include/quicr/version.h` header that is included via `include/quicr/config.h`.  Apps and the library can then use the version header as needed to get the version, description and project name. 

Updated cmake to restrict builds to linux and apple platforms as well as arm64 and x86_64.  Others will generate a fatal error indicating the platform/arch is unsupported.


README.md
```diff
@@ -6,8 +6,18 @@ libquicr
...
+
 ## Build
 
+### Dependencies
+
+#### Linux
+
+```
+sudo apt-get update && sudo apt-get install -y gcc g++ golang pkgconf cmake make 
+```
```
GhostofCookie, line 19: Might be good to add the Mac dependencies as well

TimEvens: will add that now.

GhostofCookie (approved):

PR #242, opened 2024-09-12
Several fixes to address client closes and subscribe/unsubscribe processing

### LINKED ISSUE #239: Client disconnects are not gracefully closed on in qserver
Client disconnects do not result in qserver seeing the disconnect. This causes a delay till idle timeout cleans it up.

### LINKED ISSUE #240: example qserver UnsubscribeReceived() callback is never called

### LINKED ISSUE #241: Client does not receive and process unsubscribe, publish track remains OK

### LINKED ISSUE #246: Version header is not available to parent modules
Parent cmake projects/modules that include libquicr do not automatically have access to the version header include location. 

```
In file included from /Users/tievens/Documents/workspace/ws-media10x/laps/src/config.h:6:
/Users/tievens/Documents/workspace/ws-media10x/laps/dependencies/libquicr/include/quicr/config.h:6:10: fatal error: 'quicr/version.h' file not found
#include "quicr/version.h"
         ^~~~~~~~~~~~~~~~~
In file included from /Users/tievens/Documents/workspace/ws-media10x/laps/src/main.cc:4:
In file included from /Users/tievens/Documents/workspace/ws-media10x/laps/src/config.h:6:
/Users/tievens/Documents/workspace/ws-media10x/laps/dependencies/libquicr/include/quicr/config.h:6:10: fatal error: 'quicr/version.h' file not found
#include "quicr/version.h"
         ^~~~~~~~~~~~~~~~~
1 error generated.
```

TimEvens:
Closes #239
Closes #240
Closes #241 
Closes #246


src/client.cpp
```diff
@@ -219,9 +221,15 @@ namespace quicr {
                                         "Received unsubscribe conn_id: {0} subscribe_id: {1}",
                                         conn_ctx.connection_handle,
                                         msg.subscribe_id);
-                    sub_it->second.get()->SetStatus(SubscribeTrackHandler::Status::kNotSubscribed);
 
-                    RemoveSubscribeTrack(conn_ctx, *sub_it->second);
+                    const auto pub_track_ns_it =
```
TimEvens, line 225: This is pretty ugly. While this works, I'm sure there is a much cleaner way to do the same thing.

src/transport_picoquic.cpp
```diff
@@ -325,99 +327,103 @@ PqLoopCb(picoquic_quic_t* quic, picoquic_packet_loop_cb_enum cb_mode, void* call
...
+                if (transport->pq_loop_metrics_prev_time) {
+                    transport->EmitMetrics();
                 }
 
-                if (targ->current_time - transport->pq_loop_prev_time > 100'000) {
+                transport->pq_loop_metrics_prev_time = targ->current_time;
+            }
 
-                    transport->CheckConnsForCongestion();
+            if (targ->current_time - transport->pq_loop_prev_time > 100'000) {
```
GhostofCookie, line 389: Magic number?

src/transport_picoquic.cpp
```diff
@@ -325,99 +327,103 @@ PqLoopCb(picoquic_quic_t* quic, picoquic_packet_loop_cb_enum cb_mode, void* call
...
+                picoquic_cnx_t* close_cnx = picoquic_get_first_cnx(quic);
 
+                if (close_cnx == NULL) {
                     return PICOQUIC_NO_ERROR_TERMINATE_PACKET_LOOP;
                 }
 
-                break;
+                while (close_cnx != NULL) {
+                    SPDLOG_LOGGER_INFO(
+                      transport->logger, "Closing connection id {0}", reinterpret_cast<uint64_t>(close_cnx));
```
GhostofCookie, line 412: Reinterpret cast seems strong, what type is close_cnx? Would static cast work?

src/transport_picoquic.cpp
```diff
@@ -706,9 +712,20 @@ PicoQuicTransport::Close(const TransportConnId& conn_id, uint64_t app_reason_cod
     }
 
     // Only one datagram context is per connection, if it's deleted, then the connection is to be terminated
-    OnConnectionStatus(conn_id, TransportStatus::kDisconnected);
+    switch (app_reason_code) {
+        case 1: // idle timeout
+            OnConnectionStatus(conn_id, TransportStatus::kIdleTimeout);
+            break;
+
+        case 100: // Client shutting down connection
```
GhostofCookie, line 720: Should we formalize these error codes into an enum class in another PR?

GhostofCookie (approved):

src/transport_picoquic.cpp
TimEvens: It's a pointer, so we have to reinterpret.

src/transport_picoquic.cpp
TimEvens: Yeah, we need app reason codes that we support in addition to the ones that MoQT defines. I created https://github.com/Quicr/libquicr/issues/247 to track and fix this.

CMakeLists.txt
```diff
@@ -13,12 +13,11 @@ else()
 endif()
 
 project(quicr
-        VERSION 1.0.0
+        VERSION 1.0.1
         DESCRIPTION "QuicR, a Media over QUIC Library"
         LANGUAGES CXX)
 
 configure_file( src/version.h.in ${CMAKE_BINARY_DIR}/include/quicr/version.h )
```
RichLogan, line 20: `${CMAKE_BINARY_DIR}` I think is going to be the top level build directory of whatever project is the root (which might not be libquicr). This still works because the the corresponding `target_include_directories` matches it, but it might not be what we want. We might want to be using `CMAKE_CURRENT_BINARY_DIR` instead, to keep it under libquicr. 

For example if you built a containing project (e.g laps), this would put the version header at `/someProject/build/include/quicr/version.h` not `someProject/build/dependencies/libquicr/include/quicr/version.h`

PR #251, opened 2024-09-13
Fix duplicate ready status

### LINKED ISSUE #245: Connection status is reporting ready twice

TimEvens:
Fixes #245


GhostofCookie (commented):
Looks good! Though qperf also needs the new case handled, otherwise it closes with critical, even though connection was established.

GhostofCookie (approved):

PR #257, opened 2024-09-18
Refactor uintvar

GhostofCookie:


test/uintvar.cpp
```diff
@@ -11,16 +12,36 @@ namespace var {
...
-    CHECK_EQ(4, quicr::SizeofUintV(quicr::ToUintV(var::kValue4Byte).at(0)));
-    CHECK_EQ(8, quicr::SizeofUintV(quicr::ToUintV(var::kValue8Byte).at(0)));
-}
\ No newline at end of file
+    CHECK_EQ(1, quicr::UintVar(var::kValue1Byte).Size());
+    CHECK_EQ(2, quicr::UintVar(var::kValue2Byte).Size());
+    CHECK_EQ(4, quicr::UintVar(var::kValue4Byte).Size());
+    CHECK_EQ(8, quicr::UintVar(var::kValue8Byte).Size());
+
+    CHECK_EQ(1, quicr::UintVar(var::kValue1Byte).Bytes().size());
```
TimEvens, line 28: The previous test was testing getting the size with a single byte (e.,g reading in partial data), this test requires the value to have been set, which required `UintVar(all bytes)` first.   The use-case is still there for checking the size of the uintvar by reading the first byte, which drives how many bytes are needed to load the uintvar.   This test is fine, but a new test is needed to replace what the old one was doing.

include/quicr/detail/uintvar.h
```diff
@@ -47,84 +49,66 @@ namespace quicr {
...
+        UintVar(Span<const uint8_t> bytes)
+          : _value{ 0 }
+        {
+            if (bytes.empty() || bytes.size() > sizeof(uint64_t)) {
+                throw std::invalid_argument("Invalid number of bytes for uintvar");
+            }
 
-        if (value & (0xC0ull << 56)) { // Check if invalid
-            return {};
+            std::memcpy(&_value, bytes.data(), bytes.size());
```
TimEvens, line 86: Dangerous to not use `UintVar::Size(bytes.front())` for the size.  Length could be incorrect.  For example, the data could have more trailing bytes that might be for the next uintvar.

test/uintvar.cpp
GhostofCookie: That is what the test on line 26 is handling, it just does it under the hood for you instead. This si separate test to make sure that the Bytes span size is also correct.

include/quicr/detail/uintvar.h
```diff
@@ -47,84 +49,66 @@ namespace quicr {
...
-        uint8_t* byte_value = reinterpret_cast<uint8_t*>(&value) + (sizeof(uint64_t) - uint_v.size());
-        std::memcpy(byte_value, uint_v.data(), uint_v.size());
-
-        byte_value[0] &= 0x3f; // Zero MSB length bits
+        constexpr std::size_t Size() const noexcept { return UintVar::Size(static_cast<uint8_t>(_value)); }
 
-        return SwapBytes(value);
-    }
+      private:
+        uint64_t _value;
```
TimEvens, line 112: Based on the implementation, it's a bit misleading that type of uint64_t value is not the host order (unless big-endian host) value. It's actually used for storage of the var int, which is in network byte order.

I suggest a name change here to `bytes_` with a comment  such as "Variable length integer bytes storage"

TimEvens (approved):
one nit about the storage being called `_value`

PR #258, opened 2024-09-18
Reduce URI Parse

GhostofCookie:
Uses more modern string_view find methods and doesn't use looping.

### DIFF
--- a/src/transport.cpp
+++ b/src/transport.cpp
@@ -11,55 +11,28 @@ namespace quicr {
 
     using namespace quicr::messages;
 
-    static std::optional<std::tuple<std::string, uint16_t>> ParseConnectUri(const std::string& connect_uri)
+    static std::optional<std::tuple<std::string, uint16_t>> ParseConnectUri(std::string_view connect_uri)
     {
-        // moq://domain:port/<dont-care>
         const std::string proto = "moq://";
-        auto it = std::search(connect_uri.begin(), connect_uri.end(), proto.begin(), proto.end());
-
-        if (it == connect_uri.end()) {
+        if (connect_uri.find_first_of(proto) == std::string_view::npos) {
             return std::nullopt;
         }
 
-        // move to end for moq://
-        std::advance(it, proto.length());
+        connect_uri.remove_prefix(proto.length());
 
-        std::string address_str;
-        std::string port_str;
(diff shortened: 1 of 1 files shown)


src/transport.cpp
```diff
@@ -11,56 +11,29 @@ namespace quicr {
...
+            return std::nullopt;
+        }
 
-        if (address_str.empty() || port_str.empty()) {
+        std::string address_str{ connect_uri.substr(0, colon_pos) };
+        if (address_str.empty()) {
             return std::nullopt;
         }
 
+        connect_uri.remove_prefix(colon_pos + 1);
```
TimEvens, line 33: dos this work right when colon_pos == npos (not found)?

src/transport.cpp
```diff
@@ -11,56 +11,29 @@ namespace quicr {
...
+
+        uint16_t port = std::stoi(std::string{ connect_uri });
```
TimEvens, line 35: Default should come in when port isn't specified.

src/transport.cpp
GhostofCookie: See line 24. If colon is not found, we should never come to this line.

src/transport.cpp
GhostofCookie: From my understanding of the previous code, we never allowed for port to not be specified, so this maintains that requirement.

TimEvens: That was only because of quick adding.  We do need to have a default, such as a well known port to use.

src/transport.cpp
```diff
@@ -11,56 +11,29 @@ namespace quicr {
...
-                port_str.assign(it, slash);
-                std::advance(it, port_str.length());
-                port = stoi(port_str, nullptr);
-                it++;
-                break;
-            }
-
-        } while (it != connect_uri.end());
+        auto colon_pos = connect_uri.find_first_of(':');
+        if (colon_pos == std::string_view::npos) {
```
TimEvens, line 24: URL parsing should support no port being specified.  Need to provide some well known default that we use.

TimEvens:
@GhostofCookie , keep in mind the url parsing was added quickly during the work week. It by far does not represent the complete nature of parsing the URL/URI.

PR #259, opened 2024-09-19
Enable/fix linting

RichLogan:
- Enable linting on targets & CI
- Fixup couple existing naming violations in tests


benchmark/stream_buffer.cpp
```diff
@@ -11,9 +11,9 @@
 static void
 StreamBuffer_Construct(benchmark::State& state)
 {
-    for ([[maybe_unused]] const auto& _ : state) {
-        auto __ = quicr::StreamBuffer<std::uint8_t>();
-        benchmark::DoNotOptimize(__);
+    for ([[maybe_unused]] const auto& iteration : state) {
+        auto _ = quicr::StreamBuffer<std::uint8_t>();
+        benchmark::DoNotOptimize(_);
         benchmark::ClobberMemory();
```
GhostofCookie, line 17: Probably better to maintain the state iteration as `_` and instead give a name to the temporary (which is a pattern used in the other benchmark files). The fact that it has to be marked as maybe_unused is super annoying given `_` is a common "unused" motif in other languages.

benchmark/stream_buffer.cpp
```diff
@@ -66,9 +66,9 @@ StreamBuffer_Front(benchmark::State& state)
 static void
 SafeStreamBuffer_Construct(benchmark::State& state)
 {
-    for ([[maybe_unused]] const auto& _ : state) {
-        auto __ = quicr::SafeStreamBuffer<std::uint8_t>();
-        benchmark::DoNotOptimize(__);
+    for ([[maybe_unused]] const auto& iteration : state) {
+        auto _ = quicr::SafeStreamBuffer<std::uint8_t>();
+        benchmark::DoNotOptimize(_);
```
GhostofCookie, line 71:
```suggestion
    for ([[maybe_unused]] const auto& _ : state) {
        auto buffer = quicr::SafeStreamBuffer<std::uint8_t>();
        benchmark::DoNotOptimize(buffer);
```

benchmark/stream_buffer.cpp
```diff
@@ -11,9 +11,9 @@
 static void
 StreamBuffer_Construct(benchmark::State& state)
 {
-    for ([[maybe_unused]] const auto& _ : state) {
-        auto __ = quicr::StreamBuffer<std::uint8_t>();
-        benchmark::DoNotOptimize(__);
+    for ([[maybe_unused]] const auto& iteration : state) {
+        auto _ = quicr::StreamBuffer<std::uint8_t>();
+        benchmark::DoNotOptimize(_);
```
GhostofCookie, line 16:
```suggestion
    for ([[maybe_unused]] const auto& _ : state) {
        auto buffer = quicr::StreamBuffer<std::uint8_t>();
        benchmark::DoNotOptimize(buffer);
```

benchmark/uintvar.cpp
```diff
@@ -20,9 +20,9 @@ UIntVar_FromUint64(benchmark::State& state)
 static void
 UIntVar_ToUint64(benchmark::State& state)
 {
-    constexpr auto var_int = quicr::UintVar(0x123456789);
+    constexpr auto kVarInt = quicr::UintVar(0x123456789);
```
GhostofCookie, line 23: Wow, is it nit-picky. Even for a local constexpr?

RichLogan: Theoretically I think this *is* allowed either way in the guide:

> All such variables with static storage duration (i.e., statics and globals, see Storage Duration for details) should be named this way, including those in templates where different instantiations of the template may have different values. This convention is optional for variables of other storage classes, e.g., automatic variables; otherwise the usual variable naming rules apply. For example:

But our rules aren't quite setup for that. Let me see if I can get there.

GhostofCookie: But these aren't static? It's fine, just thought it was weird it would care about a local looking like that.

RichLogan: Sorry, that's what I'm saying. These don't need the k prefix because they're not static. But our clang tidy rules that we gave setup do not match the guide quite at that level of granularity, so they're sort of giving a false positive.

GhostofCookie: Ah, makes sense. Thanks!

TimEvens (approved):

PR #260, opened 2024-09-19
Cache buffer

GhostofCookie:


benchmark/stream_buffer.cpp
```diff
@@ -63,6 +63,30 @@ StreamBuffer_Front(benchmark::State& state)
     }
 }
 
+static void
+StreamBuffer_Pop(benchmark::State& state)
+{
+    quicr::StreamBuffer<std::uint8_t> buffer;
+    std::vector<uint8_t> bytes(1'000'000'000, 0);
```
TimEvens, line 70: 1gb might be a lot for automated tests

benchmark/stream_buffer.cpp
```diff
@@ -63,6 +63,30 @@ StreamBuffer_Front(benchmark::State& state)
...
+    for ([[maybe_unused]] const auto& _ : state) {
+        buffer.Pop();
+    }
+}
+
+static void
+StreamBuffer_PopBytes(benchmark::State& state)
+{
+    quicr::StreamBuffer<std::uint8_t> buffer;
+    std::vector<uint8_t> bytes(1'000'000'000, 0);
```
TimEvens, line 82: Same, is there a reason for this value?

include/quicr/detail/cache_buffer.h
```diff
@@ -0,0 +1,102 @@
...
+#include <memory>
+#include <optional>
+#include <string>
+#include <type_traits>
+#include <vector>
+
+class CacheBuffer
+{
+    template<class It, class SpanIt>
+    class IteratorImpl
```
TimEvens, line 18: Very interesting... puts custom allocator with vector to shame...

include/quicr/detail/cache_buffer.h
```diff
@@ -0,0 +1,102 @@
...
+    friend CacheBuffer& operator<<(CacheBuffer& buffer, const Span<const uint8_t>& value)
+    {
+        buffer.Push(value);
+        return buffer;
+    }
+
+    template<typename T, std::enable_if_t<std::is_integral_v<T>, bool> = true>
+    friend CacheBuffer& operator<<(CacheBuffer& buffer, const T& value)
+    {
+        buffer.Push(Span{ reinterpret_cast<const std::uint8_t*>(&value), sizeof(T) });
```
TimEvens, line 91: Where is it restricted to data types that can support this? For example, T = map.

include/quicr/detail/cache_buffer.h
```diff
@@ -0,0 +1,102 @@
...
+
+  public:
+    using iterator = IteratorImpl<BufferType::iterator, BufferType::value_type::iterator>;
+    using const_iterator = IteratorImpl<BufferType::const_iterator, BufferType::value_type::iterator>;
+
+    CacheBuffer() = default;
+
+    void Push(Span<const uint8_t> bytes) { _buffer.push_back(std::move(bytes)); }
+
+    std::vector<uint8_t> Copy() const { return { this->begin(), this->end() }; }
```
TimEvens, line 80: If we back this storage with a shared heap or similar, we could avoid a copy. The caller can also make their own copy if they really needed it.

include/quicr/detail/cache_buffer.h
```diff
@@ -0,0 +1,102 @@
...
+class CacheBuffer
+{
+    template<class It, class SpanIt>
+    class IteratorImpl
+    {
+        using it_traits = std::iterator_traits<const std::uint8_t*>;
+        using it_impl = IteratorImpl;
+
+      public:
+        using value_type = typename it_traits::value_type;
```
TimEvens, line 24: Is there a clean way to indicate our styling is ignored here?

include/quicr/detail/cache_buffer.h
```diff
@@ -0,0 +1,102 @@
...
+        return buffer;
+    }
+
+    iterator begin() noexcept { return iterator(_buffer.begin(), _buffer.end()); }
+    iterator end() noexcept { return iterator(_buffer.end(), _buffer.end()); }
+    const_iterator begin() const noexcept { return const_iterator(_buffer.begin(), _buffer.end()); }
+    const_iterator end() const noexcept { return const_iterator(_buffer.end(), _buffer.end()); }
+
+  private:
+    BufferType _buffer;
```
TimEvens, line 101: Isn't the style to have underscore prepended?

TimEvens (commented):
Initial comments. Will run evaluation shortly

benchmark/uintvar.cpp
```diff
@@ -11,7 +11,7 @@ static void
 UIntVar_FromUint64(benchmark::State& state)
 {
     for ([[maybe_unused]] const auto& _ : state) {
-        auto value = quicr::UintVar(0x123456789);
+        auto value = quicr::UintVar(0x3FFFFFFFFFFFFFFF);
```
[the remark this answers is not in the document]
GhostofCookie: Just checking I wasn't crazy.

benchmark/stream_buffer.cpp
GhostofCookie: Because there needs to be a value to pop to test it, and the max iterations for tests (from observation only) is about that much.

include/quicr/detail/cache_buffer.h
GhostofCookie:
```cpp
template<typename T, std::enable_if_t<std::is_integral_v<T>, bool> = true>

include/quicr/detail/cache_buffer.h
GhostofCookie: That's what this is for. You don't have to copy it, you can just iterate it like a span (hence the custom iterator). If you wanted the bytes in a vector you can do this. It's a convenience method, but it doesn't have to exist.

include/quicr/detail/cache_buffer.h
GhostofCookie: I'm going to remove these, I just like to code STL style containers like STL.

include/quicr/detail/cache_buffer.h
GhostofCookie: Going to change, I just don't write code like google by hand.

TimEvens (approved):
LGTM, I like the iterator instead of Bytes.

PR #263, opened 2024-09-20
Add benchmarks for data encode/decode

TimEvens:


Makefile
```diff
@@ -17,7 +17,7 @@ all: ${BUILD_DIR}
 	cmake --build ${BUILD_DIR} --parallel 8
 
 ${BUILD_DIR}: CMakeLists.txt cmd/CMakeLists.txt
-	cmake -B${BUILD_DIR} -DBUILD_TESTING=TRUE -DQUICR_BUILD_TESTS=ON -DCMAKE_BUILD_TYPE=Debug -DUSE_MBEDTLS=OFF .
+	cmake -B${BUILD_DIR} -DBUILD_TESTING=TRUE -DQUICR_BUILD_TESTS=ON -DCMAKE_BUILD_TYPE=Release -DUSE_MBEDTLS=OFF .
```
TimEvens, line 20: Default to release if folks use `make`.

benchmark/messages.cpp
```diff
@@ -0,0 +1,111 @@
...
+static void
+Moq_EncodeGroupHeader(benchmark::State& state)
+{
+    for ([[maybe_unused]] const auto& _ : state) {
+        quicr::Serializer buffer;
+        auto hdr = quicr::messages::MoqStreamHeaderGroup{};
+        hdr.subscribe_id = 0x100;
+        hdr.track_alias = 0x100;
+        hdr.priority = 0xA;
+        hdr.group_id = 0x1;
```
GhostofCookie, line 20: Better to move this out of the loop so it doesn't get measured. Then you don't need to do the ClobberMemory thing.

GhostofCookie: Unless the intent is to see how bad the performance gets if you create these on the fly. In which case it would be good to have both tests: 1 where the buffer and message exist before the for loop, and one where they are also measured, and then we can compare.

benchmark/messages.cpp
```diff
@@ -0,0 +1,111 @@
...
+static void
+Moq_EncodeGroupObject1000Bytes(benchmark::State& state)
+{
+    const auto payload = std::vector<uint8_t>(1000, 0);
+    for ([[maybe_unused]] const auto& _ : state) {
+        quicr::Serializer buffer;
+        auto obj = quicr::messages::MoqStreamGroupObject{};
+        obj.object_id = 0x1;
+        obj.payload.reserve(payload.size());
+        obj.payload.assign(payload.begin(), payload.end());
```
GhostofCookie, line 37: Same comment as above, we probably only want to measure `buffer << obj;`.

benchmark/stream_buffer.cpp
```diff
@@ -63,6 +63,30 @@ StreamBuffer_Front(benchmark::State& state)
...
+
+static void
+StreamBuffer_PopBytes(benchmark::State& state)
+{
+    quicr::StreamBuffer<std::uint8_t> buffer;
+    std::vector<uint8_t> bytes(1'000'000, 0);
+    buffer.Push(bytes);
+
+    for ([[maybe_unused]] const auto& _ : state) {
+        buffer.Pop(10);
```
GhostofCookie, line 86: Remember this is going to pop 10 items, so for a million iterations you'll need 10 million bytes so that all calls will pop something.

TimEvens:
Includes merge of cache buffer branch, will close and open a new PR with only the benchmark changes for messages.

PR #264, opened 2024-09-24
Update tick service, remove method trace, and update metric sample time

### LINKED ISSUE #252: Change metrics sample time from time point to uint64_t in microseconds from epoch 1970

TimEvens:
* Fixes #252; Changes metrics sample time to microseconds since epoch
* Replaces method trace with ticks in microseconds
* Move tick service out of time queue header


include/quicr/detail/time_queue.h
```diff
@@ -24,26 +24,16 @@
...
-namespace quicr {
-
-    /**
-     * Interface for services that calculate ticks.
-     */
-    struct TickService
-    {
-        using TickType = size_t;
-        using DurationType = std::chrono::microseconds;
+#include "quicr/detail/tick_service.h"
```
GhostofCookie, line 34: This include could just be
```suggestion
#include "tick_service.h"
```
since they are in the same directory.

include/quicr/detail/transport.h
```diff
@@ -3,6 +3,7 @@
 
 #pragma once
 
+#include "quicr/detail/tick_service.h"
 #include <quicr/detail/messages.h>
```
GhostofCookie, line 7:
```suggestion
#include "messages.h"
#include "tick_service.h"
```

src/quic_transport.cpp
```diff
@@ -13,11 +13,13 @@ namespace quicr {
...
                                                                 const TransportConfig& tcfg,
                                                                 TransportDelegate& delegate,
+                                                                std::shared_ptr<TickService> tick_service,
                                                                 std::shared_ptr<spdlog::logger> logger)
     {
         switch (server.proto) {
             case TransportProtocol::kQuic:
-                return std::make_shared<PicoQuicTransport>(server, tcfg, delegate, false, std::move(logger));
+                return std::make_shared<PicoQuicTransport>(
+                  server, tcfg, delegate, false, tick_service, std::move(logger));
```
GhostofCookie, line 22:
```suggestion
                  server, tcfg, delegate, false, std::move(tick_service), std::move(logger));
```
To avoid the double copy.

src/quic_transport.cpp
```diff
@@ -29,12 +31,14 @@ namespace quicr {
...
                                                                 TransportDelegate& delegate,
+                                                                std::shared_ptr<TickService> tick_service,
                                                                 std::shared_ptr<spdlog::logger> logger)
     {
         switch (server.proto) {
 
             case TransportProtocol::kQuic:
-                return std::make_shared<PicoQuicTransport>(server, tcfg, delegate, true, std::move(logger));
+                return std::make_shared<PicoQuicTransport>(
+                  server, tcfg, delegate, true, tick_service, std::move(logger));
```
GhostofCookie, line 41:
```suggestion
                  server, tcfg, delegate, true, std::move(tick_service), std::move(logger));
```

src/transport.cpp
```diff
@@ -69,6 +69,7 @@ namespace quicr {
       , logger_(spdlog::stderr_color_mt("MTC"))
       , server_config_({})
       , client_config_(cfg)
+      , tick_service_(std::make_shared<ThreadedTickService>())
```
GhostofCookie, line 72: Should this be injected? Maybe this makes more sense, but wondering.

src/transport_picoquic.cpp
```diff
@@ -872,6 +868,7 @@ PicoQuicTransport::PicoQuicTransport(const TransportRemote& server,
   , serverInfo_(server)
   , delegate_(delegate)
   , tconfig_(tcfg)
+  , tick_service_(tick_service)
```
GhostofCookie, line 871:
```suggestion
  , tick_service_(std::move(tick_service))
```

src/transport.cpp
TimEvens: I was thinking that too.  client and server have access to this so each doesn't really have to create it, but it might make sense to allow client/server do it and have transport use one passed in.

GhostofCookie (approved):

GhostofCookie (approved):

PR #265, opened 2024-09-24
Add benchmark for encode/decode of data messages

TimEvens:


benchmark/messages.cpp
```diff
@@ -0,0 +1,111 @@
...
+    hdr.group_id = 0x1;
+    buffer << hdr;
+
+    const auto net_data = buffer.Take();
+    quicr::StreamBuffer<uint8_t> sbuf;
+
+    for ([[maybe_unused]] const auto& _ : state) {
+        sbuf.PushLengthBytes(net_data);
+        hdr = quicr::messages::MoqStreamHeaderGroup{};
+        sbuf >> hdr;
```
GhostofCookie, line 77: It is awkward writing benchmarks for methods that pop from a queue, but this is meauring a lot more than `sbuf >> hdr;`. There is a way to pause and resume the state timer, but that can cause some unexpected large measurement numbers. Maybe worth allocating the buffer with enough bytes for the bench that you won't have to keep pushing bytes.

benchmark/messages.cpp
```diff
@@ -26,7 +26,25 @@ Moq_EncodeGroupHeader(benchmark::State& state)
...
+Moq_EncodeGroupHeader(benchmark::State& state)
+{
+    auto hdr = quicr::messages::MoqStreamHeaderGroup{};
+    hdr.subscribe_id = 0x100;
+    hdr.track_alias = 0x100;
+    hdr.priority = 0xA;
+    hdr.group_id = 0x1;
+
+    for ([[maybe_unused]] const auto& _ : state) {
+        quicr::Serializer buffer;
```
GhostofCookie, line 38: I'd still pull the buffer out this loop just avoid measuring it, and only measure the encode. No need for the DoNotOptimize and ClobberMemory lines.

benchmark/messages.cpp
```diff
@@ -43,7 +61,26 @@ Moq_EncodeGroupObject1000Bytes(benchmark::State& state)
...
+{
+    const auto payload = std::vector<uint8_t>(1000, 0);
+
+    auto obj = quicr::messages::MoqStreamGroupObject{};
+    obj.object_id = 0x1;
+    obj.payload.reserve(payload.size());
+    obj.payload.assign(payload.begin(), payload.end());
+
+    for ([[maybe_unused]] const auto& _ : state) {
+        quicr::Serializer buffer;
```
GhostofCookie, line 74: Same as above.

benchmark/messages.cpp
```diff
@@ -57,6 +94,46 @@ Moq_EncodeGroupObject1Byte(benchmark::State& state)
...
+
+        const auto net_data = buffer.Take();
+        quicr::StreamBuffer<uint8_t> sbuf;
+
+        sbuf.PushLengthBytes(net_data);
+        hdr = quicr::messages::MoqStreamHeaderGroup{};
+        sbuf >> hdr;
+
+        benchmark::DoNotOptimize(hdr);
+        benchmark::ClobberMemory();
```
GhostofCookie, line 133: Not needed, you use the value, so should be fine without these lines.

PR #266, opened 2024-09-24
Add qperf metrics

GhostofCookie:


cmd/qperf/qperf.cpp
```diff
@@ -123,7 +130,11 @@ namespace {
             return std::shared_ptr<PerfSubscribeTrackHandler>(new PerfSubscribeTrackHandler(full_track_name));
         }
 
-        void ObjectReceived(const quicr::ObjectHeaders&, quicr::BytesSpan) override {}
+        void ObjectReceived(const quicr::ObjectHeaders&, quicr::BytesSpan data) override
+        {
+            metrics_.objects_received++;
+            metrics_.bytes_received += data.size();
```
GhostofCookie, line 136: I don't love this, but for some reason on my machine the GetMetrics for subscribers was reporting 0s.

TimEvens: It's because upon callback in transport.cpp, it's not updated. It needs to be updated under `ProcessStreamDataMessage()` and `OnRecvDgram()`

TimEvens (changes requested):
move the increments to transport.cpp

TimEvens (commented):
Looks good, but datagram needs it as well. https://github.com/Quicr/libquicr/blob/3d0672b08f27bdf9d29f994a2bc6e4657e95fa10/src/transport.cpp#L879

TimEvens (approved):
LGTM

PR #267, opened 2024-09-25
Reuse buffers

GhostofCookie:


include/quicr/detail/transport.h
```diff
@@ -327,6 +327,9 @@ namespace quicr {
 
         std::shared_ptr<ITransport> quic_transport_; // **MUST** be last for proper order of destruction
 
+        Serializer ctrl_serial_buffer_{ 4096 };
+        Serializer serial_buffer_{ 8192 };
```
GhostofCookie, line 331: These numbers should probably be formalized.

src/transport.cpp
```diff
@@ -94,7 +94,11 @@ namespace quicr {
         spdlog::drop(logger_->name());
     }
 
-    void Transport::Init() {}
+    void Transport::Init()
+    {
+        ctrl_serial_buffer_.reserve(4096);
```
TimEvens, line 99: Considering it'll grow, it's probably okay not to be in the transport config, but it would be nice to have it as a consexpr.

src/transport.cpp
```diff
@@ -94,7 +94,11 @@ namespace quicr {
...
+        serial_buffer_.reserve(8192);
```
TimEvens, line 100: Considering it'll grow, it's probably okay not to be in the transport config, but it would be nice to have it as a consexpr.

TimEvens (approved):

PR #268, opened 2024-10-03
Add esp as available platforms

BrettRegnier:


GhostofCookie (approved):

CMakeLists.txt
```diff
@@ -19,7 +19,7 @@ project(quicr
 
 configure_file( src/version.h.in ${CMAKE_BINARY_DIR}/include/quicr/version.h )
 
-IF (NOT LINUX AND NOT APPLE)
+IF (NOT LINUX AND NOT APPLE AND NOT PLATFORM_ESP_IDF)
```
GhostofCookie, line 22: And actually, while we're here it might be good to replace LINUX with UNIX since I have had this fail on Ubuntu Server 22.04 (which is, by most definitions, Linux).

BrettRegnier: I'm going to hold off on pushing this PR. There is more to add to this before we can actually get ESP working with libquicr.

BrettRegnier:
I'm going to hold off on pushing this PR. There is more to add to this before we can actually get ESP working with libquicr.

GhostofCookie:
This branch has gotten complicated and is generally hard to keep up with main. Closing it in favour of #529

PR #271, opened 2024-10-09
Fix race issue with picoquic streams and other minor updates

TimEvens:
Fixes stream race issue causing picoquic to close connections due to local error 5. 

Also added the following:

* Fixed debug logging in transport to be set when transport config enables debug
* Added reusable publish send object buffer to avoid needless allocations for every object sent

### DIFF
--- a/include/quicr/detail/priority_queue.h
+++ b/include/quicr/detail/priority_queue.h
@@ -122,7 +122,7 @@ namespace quicr {
                 if (!tqueue || tqueue->Empty())
                     continue;
 
-                return std::move(tqueue->Front());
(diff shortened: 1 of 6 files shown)


GhostofCookie (approved):
Looks good, just needs to pass formatting build.

TimEvens:
> Oh, linux is mad about a missing header `<stdexcept>` I think for those exceptions. Also looks like it's mad about some moves?

Thanks. Just made those changes. It's interesting because I didn't change those... but good to get fixed. :)

GhostofCookie (approved):

PR #282, opened 2024-10-22
Indicate remote close on remote close in connection status

TimEvens:
Fixes the connection status reason type to correctly indicate the reason for the close.

Also adds a limit of 255 byte objects for `examples/qserver`.  Example `qserver` is for simple testing and demo/examples.  Full relay functionality is only maintained under https://github.com/quicr/laps

### DIFF
--- a/cmd/examples/server.cpp
+++ b/cmd/examples/server.cpp
@@ -102,6 +102,12 @@ class MySubscribeTrackHandler : public quicr::SubscribeTrackHandler
 
     void ObjectReceived(const quicr::ObjectHeaders& object_headers, quicr::BytesSpan data) override
     {
+        if (data.size() > 255) {
+            SPDLOG_CRITICAL("Example server is for example only, received data > 255 bytes is not allowed!");
+            SPDLOG_CRITICAL("Use github.com/quicr/laps for full relay functionality");
+            throw std::runtime_error("Example server is for example only, received data > 255 bytes is not allowed!");
+        }
+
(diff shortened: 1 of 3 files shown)


cmd/examples/server.cpp
```diff
@@ -102,6 +102,12 @@ class MySubscribeTrackHandler : public quicr::SubscribeTrackHandler
 
     void ObjectReceived(const quicr::ObjectHeaders& object_headers, quicr::BytesSpan data) override
     {
+        if (data.size() > 255) {
+            SPDLOG_CRITICAL("Example server is for example only, received data > 255 bytes is not allowed!");
+            SPDLOG_CRITICAL("Use github.com/quicr/laps for full relay functionality");
+            throw std::runtime_error("Example server is for example only, received data > 255 bytes is not allowed!");
+        }
```
suhasHere, line 109: do we see any issue with allowing object sizes greater than 256 with qserver ?

suhasHere (approved):

cmd/examples/server.cpp
TimEvens: The idea is that example server should be examples only, such as simple text.  It should be less than 255 bytes in this case. You will not be able to send video because this is not intended for that.  LAPS is developed to handle video and other real-time use-cases. This is just an example of how to use the library.

GhostofCookie (approved):
Looks good, I think limiting the qserver will be good.

PR #283, opened 2024-10-25
Prefix message length for control messages.

### LINKED ISSUE #281: Support length prefixed control message

GhostofCookie:
Closes #281

### DIFF
--- a/src/messages.cpp
+++ b/src/messages.cpp
@@ -778,41 +778,77 @@ namespace quicr::messages {
         PushExtensions(buffer, msg.extensions);
         buffer << UintVar(msg.payload.size());
         buffer << msg.payload;
-
-        UintVar size(buffer.size());
-        buffer.insert(buffer.begin(), size.begin(), size.end());
         return buffer;
     }
 
     template<class StreamBufferType>
     bool operator>>(StreamBufferType& buffer, MoqObjectDatagram& msg)
     {
-        if (!ParseUintVField(buffer, msg.subscribe_id)) {
+        switch (msg.current_pos) {
+            case 0: {
+                if (!ParseUintVField(buffer, msg.subscribe_id)) {
+                    return false;
+                }
+                msg.current_pos += 1;
+                [[fallthrough]];
+            }
+            case 1: {
+                if (!ParseUintVField(buffer, msg.track_alias)) {
+                    return false;
+                }
+                msg.current_pos += 1;
+                [[fallthrough]];
+            }
+            case 2: {
+                if (!ParseUintVField(buffer, msg.group_id)) {
+                    return false;
+                }
+                msg.current_pos += 1;
+                [[fallthrough]];
+            }
+            case 3: {
+                if (!ParseUintVField(buffer, msg.object_id)) {
+                    return false;
+                }
+                msg.current_pos += 1;
+                [[fallthrough]];
+            }
+            case 4: {
+                if (!ParseUintVField(buffer, msg.priority)) {
+                    return false;
+                }
+                msg.current_pos += 1;
+                [[fallthrough]];
+            }
+            case 5: {
+                if (!ParseUintVField(buffer, msg.num_extensions)) {
+                    return false;
+                }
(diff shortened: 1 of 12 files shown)


src/messages.cpp
```diff
@@ -141,59 +137,30 @@ namespace quicr::messages {
         buffer << UintVar(msg.last_group_id);
         buffer << UintVar(msg.last_object_id);
 
+        UintVar size(buffer.size());
+        buffer.insert(buffer.begin(), size.begin(), size.end());
+
```
suhasHere, line 142: wouldn't this causes reshuffling of the elements ?

src/messages.cpp
```diff
@@ -972,77 +778,41 @@ namespace quicr::messages {
         PushExtensions(buffer, msg.extensions);
         buffer << UintVar(msg.payload.size());
         buffer << msg.payload;
+
+        UintVar size(buffer.size());
+        buffer.insert(buffer.begin(), size.begin(), size.end());
```
suhasHere, line 783: Non control message doesn't need length update

src/messages.cpp
```diff
@@ -972,77 +778,41 @@ namespace quicr::messages {
...
         return buffer;
     }
 
     template<class StreamBufferType>
     bool operator>>(StreamBufferType& buffer, MoqObjectDatagram& msg)
     {
-        switch (msg.current_pos) {
```
suhasHere, line 981: ObjectDatagram is not a control message

src/messages.cpp
```diff
@@ -1056,40 +826,22 @@ namespace quicr::messages {
         buffer << UintVar(msg.subscribe_id);
         buffer << UintVar(msg.track_alias);
         buffer << UintVar(msg.priority);
+
+        UintVar size(buffer.size());
+        buffer.insert(buffer.begin(), size.begin(), size.end());
```
suhasHere, line 831: Not a control message

src/messages.cpp
```diff
@@ -1056,40 +826,22 @@ namespace quicr::messages {
...
         return buffer;
     }
 
     template<class StreamBufferType>
     bool operator>>(StreamBufferType& buffer, MoqStreamHeaderTrack& msg)
```
suhasHere, line 836: Not a control message:

test/moq_messages.cpp
```diff
@@ -58,6 +58,23 @@ Verify(std::vector<uint8_t>& buffer, uint64_t message_type, T& message, [[maybe_
...
+    StreamBuffer<uint8_t> in_buffer;
+    in_buffer.Push(buffer);
+
+    auto length = in_buffer.DecodeUintV();
+    CHECK(length.has_value());
+    CHECK_EQ(*length, in_buffer.Size());
+
+    auto msg_type = in_buffer.DecodeUintV();
+    CHECK(msg_type.has_value());
+    CHECK_EQ(*msg_type, message_type);
```
suhasHere, line 74: I think the order is wrong here .
Its Type, Length and body in that order.

suhasHere (changes requested):
Thanks for the PR;

- This PR updates non control messages. Non control messages are not updated for length prefix
- Tests seems to look at wrong order

On something nice to do , but can wait is  to avoid reshuflfing memory by doing
- assume length can be encoded in 2 bytes and process the buffer. 
- if length exceed 2 bytes, then reallocate the buffer.
Most of the control message lengths defined today can be defined in 2 bytes.

src/messages.cpp
GhostofCookie: Yep! Thank UintVar for that.

TimEvens: Would have been nice to be fixed value instead of uintvar for header length.  @suhasHere , do you have another way to do it?

TimEvens (commented):
Had meeting with Tomas. Changes are coming that should address the questions as well as improve processing using header length.

TimEvens (approved):
I really like the changes.

suhasHere (approved):

Issue #286, opened 2024-10-26
support for sub groups


suhasHere:
> 

- [x] SubGroup Message Encode/Decode
- [x] SubGroup Stream Mapping
- [x] Add example showing group with single sub-group, N sub-groups , < N sub-groups
- [x] Update StreamPerObject and StreaPerGroup flows
-

GhostofCookie:
Is #287 still needed if #289 exists?

PR #287, opened 2024-10-26
SubGroup message encode/code

suhasHere:
Addresses first part of #286

### DIFF
--- a/src/messages.cpp
+++ b/src/messages.cpp
@@ -1276,6 +1276,122 @@ namespace quicr::messages {
     template bool operator>> <StreamBuffer<uint8_t>>(StreamBuffer<uint8_t>&, MoqStreamGroupObject&);
     template bool operator>> <SafeStreamBuffer<uint8_t>>(SafeStreamBuffer<uint8_t>&, MoqStreamGroupObject&);
 
+
+    Bytes& operator<<(Bytes& buffer, const MoqStreamHeaderSubGroup& msg)
+    {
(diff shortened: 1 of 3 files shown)


src/messages.cpp
```diff
@@ -1276,6 +1276,122 @@ namespace quicr::messages {
...
+            }
+            case 2: {
+                if (!ParseUintVField(buffer, msg.subgroup_id)) {
+                    return false;
+                }
+                msg.current_pos += 1;
+                [[fallthrough]];
+            }
+            case 3: {
+                if (!ParseUintVField(buffer, msg.priority)) {
```
TimEvens, line 1316: This value is called out as 8 bits in the spec and is not a UintVar field. While parsing it would work in most cases, it will introduce parsing errors if priority is greater than 127.

src/messages.cpp
```diff
@@ -1276,6 +1276,122 @@ namespace quicr::messages {
...
+        switch (msg.current_pos) {
+            case 0: {
+                if (!ParseUintVField(buffer, msg.object_id)) {
+                    return false;
+                }
+                msg.current_pos += 1;
+                [[fallthrough]];
+            }
+            case 1: {
+                if (!ParseUintVField(buffer, msg.num_extensions)) {
```
TimEvens, line 1358: How is object status handled here when 1 may be payload length or object status? 

Extensions seems to be missing from the spec... but it looks like it should be considered the head of payload.

TimEvens (changes requested):

PR #289, opened 2024-10-28
Support subgroup streams 

suhasHere:
Clean up MOQStreamObject/MoQStreamGroup
Support for subgroups
update example to use subgroups


suhasHere:
Addresses #286

cmd/examples/client.cpp
```diff
@@ -175,12 +176,18 @@ DoPublisher(const quicr::FullTrackName& full_track_name, const std::shared_ptr<q
...
         }
+
         if (object_id % 10 == 0) { // Set new group
             object_id = 0;
+            subgroup_id = 0;
             group_id++;
         }
 
-        quicr::ObjectHeaders obj_headers = { group_id,       object_id++,  msg.size(),  2 /*priority*/,
+        if (object_id == 4) {
```
GhostofCookie, line 186: Magic number?

cmd/examples/server.cpp
```diff
@@ -102,6 +102,7 @@ class MySubscribeTrackHandler : public quicr::SubscribeTrackHandler
 
     void ObjectReceived(const quicr::ObjectHeaders& object_headers, quicr::BytesSpan data) override
     {
+
```
GhostofCookie, line 105: Why the newline?

include/quicr/detail/base_track_handler.h
```diff
@@ -19,9 +19,8 @@ namespace quicr {
     enum class TrackMode : uint8_t
     {
         kDatagram,
-        kStreamPerObject,
-        kStreamPerGroup,
-        kStreamPerTrack
+        kStreamPerTrack,
+        kStreamPerSubGroup,
```
GhostofCookie, line 23: No more group? I thought `kStreamPerTrack` was the one going the way of the dodo?

include/quicr/detail/messages.h
```diff
@@ -64,6 +65,7 @@ namespace quicr::messages {
 
         STREAM_HEADER_TRACK = 0x50,
         STREAM_HEADER_GROUP,
+        STREAM_HEADER_SUBGROUP = 0x4,
```
GhostofCookie, line 68: Prefer keeping the 2 digit format of these numbers for clarity (0x0 being the exception)
```suggestion
        STREAM_HEADER_SUBGROUP = 0x04,
```

include/quicr/object.h
```diff
@@ -18,8 +18,9 @@ namespace quicr {
      */
     struct ObjectHeaders
     {
-        uint64_t group_id;                   ///< Object group ID - Should be derived using time in microseconds
-        uint64_t object_id;                  ///< Object ID - Start at zero and increment for each object in group
+        uint64_t group_id;  ///< Object group ID - Should be derived using time in microseconds
+        uint64_t object_id; ///< Object ID - Start at zero and increment for each object in group
+        std::optional<uint64_t> subgroup_id;
```
GhostofCookie, line 23: Should we add a description to keep with standard?

src/publish_track_handler.cpp
```diff
@@ -36,25 +36,18 @@ namespace quicr {
...
-                if (not sent_track_header_) {
-                    is_stream_header_needed = true;
-                    sent_track_header_ = true;
-                }
-                break;
+
+        if (prev_sub_group_id_ != object_headers.subgroup_id) {
+            is_stream_header_needed = true;
+        } else if (prev_object_group_id_ != object_headers.group_id) {
+            is_stream_header_needed = true;
```
GhostofCookie, line 43: These could be compressed to 1 if condition to reduce branches.

src/transport.cpp
```diff
@@ -599,69 +599,55 @@ namespace quicr {
...
-            }
 
-            case TrackMode::kStreamPerGroup: {
+            case TrackMode::kStreamPerTrack: {
                 eflags.use_reliable = true;
 
                 if (stream_header_needed) {
                     eflags.new_stream = true;
-                    eflags.clear_tx_queue = true;
-                    eflags.use_reset = true;
```
GhostofCookie, line 625: I'm unsure about this, but I also have no clue if it was/is needed. @TimEvens ?

src/transport.cpp
```diff
@@ -1306,29 +1247,13 @@ namespace quicr {
...
       std::shared_ptr<SafeStreamBuffer<uint8_t>>&);
 
-    template std::pair<messages::MoqObjectStream&, bool> Transport::ParseDataMessage<messages::MoqObjectStream>(
-      std::shared_ptr<SafeStreamBuffer<uint8_t>>& stream_buffer,
-      MoqMessageType msg_type);
-
-    template std::pair<messages::MoqStreamHeaderTrack&, bool>
-    Transport::ParseStreamData<messages::MoqStreamHeaderTrack, messages::MoqStreamTrackObject>(
-      std::shared_ptr<SafeStreamBuffer<uint8_t>>& stream_buffer,
-      MoqMessageType msg_type);
```
GhostofCookie, line 1329: This cannot be deleted, it is still used on line 1048

GhostofCookie (changes requested):

cmd/examples/client.cpp
suhasHere: yes for the test app :-) we have group of 10 objects and i was thinking of creating an extra stream after 4 objects

src/transport.cpp
```diff
@@ -1143,6 +1085,7 @@ namespace quicr {
                     handler->subscribe_track_metrics_.bytes_received += obj.payload.size();
                     handler->ObjectReceived({ obj.group_id,
                                               obj.object_id,
+                                              std::nullopt,
```
TimEvens, line 1088: Same comment as before. Starting to get confusing for the application implementer on which object headers are supported based on what they publish that doesn't result in objects being received with those...

include/quicr/detail/base_track_handler.h
suhasHere: yes .. stream per group and stream per object gets subsumed by subgroups

TimEvens (changes requested):

src/publish_track_handler.cpp
```diff
@@ -36,25 +36,14 @@ namespace quicr {
...
-            case TrackMode::kStreamPerTrack:
-                if (not sent_track_header_) {
-                    is_stream_header_needed = true;
-                    sent_track_header_ = true;
-                }
-                break;
+
+        // change in subgroups and groups require a new stream
+        if (prev_sub_group_id_ != object_headers.subgroup_id || prev_object_group_id_ != object_headers.group_id) {
+            is_stream_header_needed = true;
```
GhostofCookie, line 42: Could collapse this to be simply
```cpp
bool is_stream_header_needed = prev_sub_group_id_ != object_headers.subgroup_id || prev_object_group_id_ != object_headers.group_id;
```
Then there would be no branches at all!

GhostofCookie (approved):
Looks good! Just needs to resolve conflicts with main.

PR #297, opened 2024-10-29
Ad support for Object Status encode/decode

suhasHere:
#295

### DIFF
--- a/include/quicr/object.h
+++ b/include/quicr/object.h
@@ -14,7 +14,8 @@ namespace quicr {
     /**
      * @brief Status of object as reported by the publisher
      */
-    enum struct ObjectStatus: uint8_t {
+    enum struct ObjectStatus : uint8_t
+    {
         AVAIALABLE = 0x0,
         DOES_NOT_EXIST = 0x1,
         END_OF_GROUP = 0x3,
--- a/include/quicr/object.h
+++ b/include/quicr/object.h
@@ -12,6 +12,17 @@ namespace quicr {
     using Extensions = std::map<uint64_t, std::vector<uint8_t>>;
 
     /**
+     * @brief Status of object as reported by the publisher
+     */
+    enum struct ObjectStatus: uint8_t {
+        AVAIALABLE = 0x0,
+        DOES_NOT_EXIST = 0x1,
+        END_OF_GROUP = 0x3,
+        END_OF_TRACK = 0x4,
+        END_OF_SUBGROUP = 0x5
+    };
+
+    /**
      * @brief Object headers struct
      *
      * @details Object headers are passed when sending and receiving an object. The object headers describe the object.
@@ -22,6 +33,7 @@ namespace quicr {
         uint64_t object_id;                  ///< Object ID - Application defined order of generation
         uint64_t subgroup_id{ 0 };           ///< Subgroup ID - Starts at 0, monotonically increases by 1
         uint64_t payload_length;             ///< Length of payload of the object data
+        ObjectStatus status;                 ///< Status of the object at the publisher
         std::optional<uint8_t> priority;     ///< Priority of the object, lower value is better
         std::optional<uint16_t> ttl;         ///< Object time to live in milliseconds
         std::optional<TrackMode> track_mode; ///< Track Mode of how the object was received or mode to use when sending
--- a/include/quicr/object.h
+++ b/include/quicr/object.h
@@ -16,11 +16,11 @@ namespace quicr {
      */
     enum struct ObjectStatus : uint8_t
     {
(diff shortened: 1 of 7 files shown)


include/quicr/object.h
```diff
@@ -11,6 +11,18 @@
...
+    /**
+     * @brief Status of object as reported by the publisher
+     */
+    enum struct ObjectStatus : uint8_t
+    {
+        AVAIALABLE = 0x0,
+        DOES_NOT_EXIST = 0x1,
+        END_OF_GROUP = 0x3,
+        END_OF_TRACK = 0x4,
+        END_OF_SUBGROUP = 0x5
```
GhostofCookie, line 23: I think naming enums like this goes against the styling.
```suggestion
        kAvailable = 0x0,
        kDoesNotExist = 0x1,
        kEndOfGroup = 0x3,
        kEndOfTrack = 0x4,
        kEndOfSubgroup = 0x5
```

src/messages.cpp
```diff
@@ -31,6 +31,12 @@ namespace quicr::messages {
                                 std::optional<Extensions>& extensions,
                                 std::optional<uint64_t>& current_tag)
     {
+
+        // get the count of extensions
+        if (!count && !ParseUintVField(buffer, count)) {
```
GhostofCookie, line 36: Prefer not coercing integer type into bool
```suggestion
        if (count == 0 && !ParseUintVField(buffer, count)) {
```

src/messages.cpp
```diff
@@ -881,8 +887,16 @@ namespace quicr::messages {
...
+        if (!msg.payload.size()) {
+            // empty payload needs a object status to be set
+            buffer << UintVar(msg.payload.size());
+            buffer << UintVar(static_cast<uint8_t>(msg.object_status));
+            PushExtensions(buffer, msg.extensions);
+        } else {
+            buffer << UintVar(msg.payload.size());
+            PushExtensions(buffer, msg.extensions);
+            buffer << msg.payload;
+        }
```
GhostofCookie, line 899: Payload size is going to be written either way:
```suggestion
        buffer << UintVar(msg.payload.size());
        if (!msg.payload.size()) {
            // empty payload needs a object status to be set
            buffer << UintVar(static_cast<uint8_t>(msg.object_status));
            PushExtensions(buffer, msg.extensions);
        } else {
            PushExtensions(buffer, msg.extensions);
            buffer << msg.payload;
        }
```

src/messages.cpp
```diff
@@ -1037,15 +1070,23 @@ namespace quicr::messages {
...
```
GhostofCookie, line 1082: Payload size will be written anyways
```suggestion
        buffer << UintVar(msg.payload.size());
        if (msg.payload.empty()) {
            // empty payload needs a object status to be set
            buffer << UintVar(static_cast<uint8_t>(msg.object_status));
            PushExtensions(buffer, msg.extensions);
        } else {
            PushExtensions(buffer, msg.extensions);
            buffer << msg.payload;
        }
```

src/messages.cpp
GhostofCookie: Also, prefer using `std::vector::empty` instead of coercing size_t to bool:
```cpp
if (msg.payload.empty()) {

Comment (GhostofCookie) [reply src/messages.cpp:899]: Something else just occurred to me; Why bother checking the payload length, if the status flag is already set to "Does not Exist"? We could just check the status. And to that end, if we set "Does not Exist", why even bother checking the payload at all? Could just ignore it entirely (even if there is stuff in it for some weird reason). Should that status be set here instead, to help reduce us manually setting the flag? If the payload is empty then it follows that the status would be set to "Does not exist".

File: test/moq_messages.cpp [hunk 5]
```diff
@@ -552,7 +552,12 @@ ObjectDatagramEncodeDecode(bool extensions)
     object_datagram.object_id = 0xFF;
     object_datagram.priority = 0xA;
     object_datagram.extensions = extensions ? kOptionalExtensions : std::nullopt;
-    object_datagram.payload = { 0x1, 0x2, 0x3, 0x5, 0x6 };
+    if (empty_payload) {
+        object_datagram.object_status = quicr::ObjectStatus::DOES_NOT_EXIST;
+        object_datagram.payload = {};
+    } else {
+        object_datagram.payload = { 0x1, 0x2, 0x3, 0x5, 0x6 };
+    }
```
GhostofCookie, line 560: Something about manually setting this status and then having to even care about the payload contents seems odd... Shouldn't it be one or the other? If the status is DOES_NOT_EXIST then who cares what's in the payload?

GhostofCookie (approved):

PR #298, opened 2024-10-29
Changing namespace to be a tuple.

### LINKED ISSUE #290: name spaces as tuple
@GhostofCookie   I think, we need following sub items to finish this work. 
- [ ]  Tuple Type support for Track Namespaces
- [ ]   Move TrackNamespace to Tuple Type enc/dec
- [ ]  Add support for the same in subscribe flows
- [ ]  Subscribe Announces
- [ ]  Support namespace as tuples within state-machine , aliases and so on
- [ ]   Match full track names against namespace tuples

GhostofCookie:
Closes #290

### DIFF
--- a/include/quicr/track_name.h
+++ b/include/quicr/track_name.h
@@ -104,8 +104,8 @@ namespace quicr {
         }
 
         TrackNamespace(const TrackNamespace& other)
-          : bytes_{ other.bytes_ }
-          , entries_{ other.entries_ }
+          : bytes_(other.bytes_)
+          , entries_(other.entries_)
         {
             std::size_t offset = 0;
             std::size_t i = 0;
@@ -116,8 +116,8 @@ namespace quicr {
         }
 
         TrackNamespace(TrackNamespace&& other)
-          : bytes_{ std::move(other.bytes_) }
-          , entries_{ other.entries_ }
+          : bytes_(std::move(other.bytes_))
+          , entries_(std::move(other.entries_))
         {
             other.entries_.clear();
 
@@ -147,8 +147,7 @@ namespace quicr {
         TrackNamespace& operator=(TrackNamespace&& other)
         {
             this->bytes_ = std::move(other.bytes_);
-            this->entries_ = other.entries_;
-            other.entries_.clear();
+            this->entries_ = std::move(other.entries_);
 
             std::size_t offset = 0;
             std::size_t i = 0;
--- a/include/quicr/track_name.h
+++ b/include/quicr/track_name.h
@@ -3,12 +3,119 @@
 
 #pragma once
 
+#include "detail/span.h"
+
+#include <algorithm>
 #include <cstdint>
 #include <optional>
+#include <tuple>
 #include <vector>
 
 namespace quicr {
-    using TrackNamespace = std::vector<uint8_t>;
+    class TrackNamespace
+    {
+      public:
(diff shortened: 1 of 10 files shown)


src/messages.cpp
```diff
@@ -142,6 +142,40 @@ namespace quicr::messages {
...
+    {
+        const auto& entries = ns.GetEntries();
+
+        Bytes payload;
+        for (const auto& entry : entries) {
+            payload << UintVar(entry.size());
+            payload << entry;
+        }
+
+        buffer << UintVar(payload.size());
```
suhasHere, line 155: this should be the following on the wire
Varint(num_entries) entry1, entry2, ... 

``` buffer << UintVar(entries.size()) ````

File: include/quicr/track_name.h [hunk 2]
```diff
@@ -3,12 +3,110 @@
...
+
+        friend bool operator<(const TrackNamespace& lhs, const TrackNamespace& rhs) noexcept
+        {
+            return lhs.bytes_ < rhs.bytes_;
+        }
+
+        friend bool operator>(const TrackNamespace& lhs, const TrackNamespace& rhs) noexcept
+        {
+            return lhs.bytes_ > rhs.bytes_;
+        }
```
suhasHere, line 87: this is useful. can you please add few tests for comparing different things ( like you had for qnames) .. basically some form of prefix tests

suhasHere (commented):
This is looking good. 
One comment on fixing the serialization and one more on adding few tests for  matching various prefixes of different sizes.

Few examples

(“example.com”, “chat555, “user1”, “dev1”, “time1”)
(“example.com”, “chat555”, “user1”, “dev2”, “time1”)
(“example.com”, “chat555”, “user1”, “dev1”, “time3”)
(“example.com”, “chat555”, “user2, “dev1”, “time4”)

==> when queried for 
(“example.com”, “chat555”) returns all the 4 entries

==> when queried for 
(“example.com”, “chat555”) returns all the 4 entries

==> when queried for 
(“example.com”, “chat555”, "user1") returns 3 entries

==> when queried for 
(“example.com”, “chat555”, "user2") returns 1 entry

and

==> when queried for 
(“example.com”, “cha123”) returns nothing

src/messages.cpp
GhostofCookie: I had started with that, but then didn't need it. I can add it back.

include/quicr/track_name.h
GhostofCookie: I'm not fully sure as of yet whether or not it does the kind of prefix checking we want, but it may work well for it. I'll add the tests.

src/messages.cpp
suhasHere: i can see that . The spec says this 

x(tuple):
Indicates that x is a tuple, consisting of a variable length integer encoded , followed by that many variable length tuple fields, each of which are encoded as variable length integer encoding, followed by that many bytes of binary data

include/quicr/track_name.h
suhasHere: I have few thoughts that I would like to go through you tomorrow on matching at tuple boundaries. This might need adding boundary information in tuple elements.

suhasHere:
Also can we update client/server example to show using namespace tuple

suhasHere (approved):

PR #303, opened 2024-10-31
Support Subscribe Announces 

suhasHere:


src/client.cpp
```diff
@@ -299,6 +299,37 @@ namespace quicr {
...
+                }
+                return true;
+            }
+            case messages::MoqMessageType::ANNOUNCE: {
+                messages::MoqAnnounce msg;
+                msg_bytes >> msg;
+
+                SPDLOG_LOGGER_DEBUG(logger_, "Received announce on conn_id: {0}", conn_ctx.connection_handle);
+
+                for (const auto& entry : conn_ctx.announce_subscriptions) {
```
GhostofCookie, line 322: Prefer structured bindings to avoid using hard to read `first` and `second`.

src/messages.cpp
```diff
@@ -675,6 +675,110 @@ namespace quicr::messages {
         return buffer >> msg.new_session_uri;
     }
 
+    //
+    // Subscribe Announces
+    //
+
+    //
+    // Announce
+    //
```
GhostofCookie, line 684: Double comment section

src/server.cpp
```diff
@@ -29,6 +29,12 @@ namespace quicr {
 
     void Server::ResolveAnnounce(ConnectionHandle, const TrackNamespace&, const AnnounceResponse&) {}
 
+    void Server::SubscribeAnnouncesReceived(ConnectionHandle connection_handle, TrackNamespace& track_namespace_prefix)
```
GhostofCookie, line 32: Remove unused names to avoid compiler complaining.

src/transport.cpp
```diff
@@ -490,6 +520,47 @@ namespace quicr {
...
+            return;
+        }
+
+        auto it = conn_it->second.announce_subscriptions.find(ns_prefix);
+        if (it != conn_it->second.announce_subscriptions.end()) {
+            // do nothing
+            return;
+        }
+
+        conn_it->second.announce_subscriptions[ns_prefix] = std::move(handler);
```
GhostofCookie, line 545: `it` already is `conn_it->second.announce_subscriptions[ns_prefix]`, we don't need to look it up again. In map this is effectively an O(1) operation though, so maybe it doesn't matter.

include/quicr/detail/messages.h
```diff
@@ -59,6 +59,10 @@ namespace quicr::messages {
 
         GOAWAY = 0x10,
 
+        SUBSCRIBE_ANNOUNCES = 0x11,
```
TimEvens, line 62: All of these need to be style formatted. For example `kSubscribeAnnounces`

include/quicr/subscribe_announces_handler.h
```diff
@@ -0,0 +1,94 @@
...
+        TrackNamespace GetTrackNamespacePrefix() const noexcept { return track_namespace_prefix_; }
+
+        // --------------------------------------------------------------------------
+        // Public Virtual API callback event methods
+        // --------------------------------------------------------------------------
+        /** @name Callbacks
+         */
+        ///@{
+
+        virtual void MatchingTrackNamespaceReceived([[maybe_unused]] const quicr::TrackNamespace& track_namespace) {}
```
TimEvens, line 57: name change

src/client.cpp
```diff
@@ -302,6 +302,37 @@ namespace quicr {
                 conn_ctx.setup_complete = true;
                 return true;
             }
+            case messages::ControlMessageType::SUBSCRIBE_ANNOUNCES_OK: {
```
TimEvens, line 305: `kSubscribeAnnouncesOk`

src/client.cpp
```diff
@@ -302,6 +302,37 @@ namespace quicr {
...
+                  logger_, "Received subscribe_announce ok, conn_id: {0}", conn_ctx.connection_handle);
+
+                auto it = conn_ctx.announce_subscriptions.find(msg.track_namespace_prefix);
+                if (it->second.get()->GetStatus() != SubscribeAnnouncesHandler::Status::kOk) {
+                    // TODO : report status to the handler
+                    it->second.get()->SetStatus(SubscribeAnnouncesHandler::Status::kOk);
+                }
+                return true;
+            }
+            case messages::ControlMessageType::ANNOUNCE: {
```
TimEvens, line 319: `kAnnounce`

src/server.cpp
```diff
@@ -111,10 +118,6 @@ namespace quicr {
 
                 // TODO(tievens): add filter type when caching supports it
                 SubscribeReceived(conn_ctx.connection_handle, msg.subscribe_id, msg.track_alias, tfn, {});
-
-                // TODO(tievens): Delay the subscribe OK till ResolveSubscribe() is called
-                SendSubscribeOk(conn_ctx, msg.subscribe_id, kSubscribeExpires, false);
```
TimEvens, line 116: Removing this would require it to be added to resolve subscribe, right?  I don't see that changed added.

cmd/examples/server.cpp
```diff
@@ -86,6 +93,15 @@ namespace qserver_vars {
     std::unordered_map<quicr::messages::TrackAlias,
                        std::unordered_map<quicr::ConnectionHandle, std::shared_ptr<quicr::SubscribeTrackHandler>>>
       pub_subscribes;
+
+    static std::string to_ascii(const std::vector<Span<const uint8_t>>& data)
+    {
+        std::stringstream output;
+        for (const auto& entry : data) {
+            output << entry.data() << ",";
+        }
```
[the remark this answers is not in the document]
GhostofCookie: But what is the format? Commas? Slashes? Colons? Does it matter and we can just pick one? Wasn't specified, so wasn't added.

include/quicr/detail/messages.h
suhasHere: yes I saw the incosistency. I feel that can be done in a follow up PR.

src/server.cpp
suhasHere: good catch . That was error . I will add it back

TimEvens (commented):
Looks good, but doesn't seem to be working correctly... likely because `Contains()` is being used to match the track namespace prefix.  

For example, subscribe to announces `a,b`... Publish announces of `a,b` works as expected, but does not work for `a,b,?` and `a,b,**`


There isn't a debug message for the announce subscribe, but I did announce subscribe to `x,y` and then published announce using `x,y,z` which resulted in the below log message.

```
[2024-11-01 15:18:59.277] [info] [server.cpp:349] No match Found for announced namespace. prefix: x/y, namespace: x/y
```

Using anything other than `x,y` for the announce, it doesn't match.  If I publish announce using `x,y` it does work with:

```
[2024-11-01 15:22:24.285] [info] [server.cpp:342] Found a match for announced namespace. prefix: x/y, namespace: x/y
```

So in other words, prefix matching is not working for announce subscribes.

TimEvens:
Many changes since this PR.  Issues #409, #378, #358 will address this.

PR #310, opened 2024-11-01
cmake can't recognize container build environment as Linux

tyqu-cisco:
Webex services will build this repo inside build container, but cmake can't recognize it as Linux, Unix is fine.


TimEvens (approved):
Thanks for submitting this. This was supposed to have been updated a while ago.

PR #311, opened 2024-11-01
expose connection handle to external application

tyqu-cisco:
For one homer server, it should be able to connect different relay servers for the different conferences, this connection handle should be defined as one map identification in libquicr.

### DIFF
--- a/include/quicr/client.h
+++ b/include/quicr/client.h
@@ -220,6 +220,16 @@ namespace quicr {
             }
         }
 
+        /**
+         * @brief Get the connection handle
+         *
+         * @return ConnectionHandle of the client
+         */
+        std::optional<ConnectionHandle> GetConnectionHandle() const
+        {
+            return connection_handle_;
+        }
+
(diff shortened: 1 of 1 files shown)


TimEvens (approved):
This is fine for now,  but we are planning to add support for multiple client connections. This will change in the near future where a client can have multiple connection handles/ids.

suhasHere:
@tyqu-cisco  can you please run `make format` once and submit the updated change.

tyqu-cisco:
> @tyqu-cisco can you please run `make format` once and submit the updated change.

Different repos have the different rules, :)
Fixed, and all checks have passed.

tyqu-cisco:
> This is fine for now, but we are planning to add support for multiple client connections. This will change in the near future where a client can have multiple connection handles/ids.

From Webex side, this handle will help us identify the different connection, which will send one specific stream to relay server. Multiple connections should be better for Webex applications.

PR #319, opened 2024-11-07
Capture publish handler weakly

RichLogan:
I believe the track handler should be captured weakly into the publish object function in order for it to not hang around after being unpublished. 

I think I proved this with a `shared_ptr::use count` test that looks like: 

```cpp
auto handler = TestPublishTrackHandler::Create();
auto config = quicr::ClientConfig();
config.connect_uri = "moq://127.0.0.1:1234";
auto client = quicr::Client(config);
client.Connect();
CHECK_EQ(handler.use_count(), 1); // Just us, the creator.
client.PublishTrack(handler);
CHECK_GT(handler.use_count(), 1); // Us, plus whatever in the client holds onto it.
client.UnpublishTrack(handler);
CHECK_EQ(handler.use_count(), 1); // Without this fix, this would be 2. 
  ```

but I don't think it's a reliable enough test to add without a framework to either a mock a client or have a in-memory server available to the tests - I didn't want to go off the deep end in this PR. 

I also removed the unused subscribe ID capture that was there.

### DIFF
--- a/src/transport.cpp
+++ b/src/transport.cpp
@@ -575,17 +575,22 @@ namespace quicr {
                                              false);
 
         // Setup the function for the track handler to use to send objects with thread safety
-        track_handler->publish_object_func_ = [&, track_handler, subscribe_id = track_handler->GetSubscribeId()](
-                                                uint8_t priority,
-                                                uint32_t ttl,
-                                                bool stream_header_needed,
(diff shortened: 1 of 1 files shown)


GhostofCookie (approved):

TimEvens (approved):
LGTM

PR #321, opened 2024-11-08
Fix subscribe compare and add missing subscribe attributes

TimEvens:
Subscribe priority and order was not passed up to the server. This commit adds that.

### DIFF
--- a/cmd/examples/server.cpp
+++ b/cmd/examples/server.cpp
@@ -56,26 +56,28 @@ namespace qserver_vars {
      * Map of subscribes set by namespace and track name hash
      *      Set<subscribe_who> = subscribe_active[track_namespace_hash][track_name_hash]
      */
-    struct SubscribeWho
+    struct SubscribeInfo
     {
         uint64_t connection_handle;
         uint64_t subscribe_id;
         uint64_t track_alias;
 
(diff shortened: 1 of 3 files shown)


cmd/examples/server.cpp
```diff
@@ -56,26 +56,28 @@ namespace qserver_vars {
...
     {
         uint64_t connection_handle;
         uint64_t subscribe_id;
         uint64_t track_alias;
 
-        bool operator<(const SubscribeWho& other) const
+        bool operator<(const SubscribeInfo& other) const
         {
-            return connection_handle < other.connection_handle && subscribe_id << other.subscribe_id;
+            return connection_handle < other.connection_handle ||
```
TimEvens, line 67: @shenning00  found this out where we weren't removing state. crazy that this wasn't caught sooner.

include/quicr/common.h
```diff
@@ -7,6 +7,9 @@
 #include <string>
 
 namespace quicr {
+    namespace messages {
+        enum struct GroupOrder : uint8_t;
```
TimEvens, line 11: Forward because messages includes common.h.

GhostofCookie (approved):

PR #325, opened 2024-11-12
Implementing Fetch

GhostofCookie:
Related to https://github.com/Quicr/laps/issues/70

### DIFF
--- a/include/quicr/detail/transport.h
+++ b/include/quicr/detail/transport.h
@@ -286,7 +286,15 @@ namespace quicr {
                        messages::ObjectPriority priority,
                        messages::GroupOrder group_order,
                        messages::GroupId start_group,
-                       messages::GroupId end_group);
+                       messages::GroupId start_object,
+                       messages::GroupId end_group,
+                       messages::GroupId end_object);
+        void SendFetchOk(ConnectionContext& conn_ctx,
+                         uint64_t subscribe_id,
+                         messages::GroupOrder group_order,
+                         bool end_of_track,
+                         messages::GroupId largest_group_id,
+                         messages::GroupId largest_object_id);
         void SendFetchError(ConnectionContext& conn_ctx,
                             uint64_t subscribe_id,
                             messages::FetchError error,
--- a/include/quicr/detail/transport.h
+++ b/include/quicr/detail/transport.h
@@ -12,6 +12,7 @@
 #include <chrono>
 #include <quicr/common.h>
 #include <quicr/config.h>
+#include <quicr/fetch_track_handler.h>
 #include <quicr/metrics.h>
 #include <quicr/publish_track_handler.h>
 #include <quicr/subscribe_track_handler.h>
@@ -142,10 +143,18 @@ namespace quicr {
          * @brief Unpublish track
          *
          * @param connection_handle           Connection ID from transport for the QUIC connection context
-         * @param track_handler    Track handler used when published track
+         * @param track_handler               Track handler used when published track
          */
         void UnpublishTrack(ConnectionHandle connection_handle,
                             const std::shared_ptr<PublishTrackHandler>& track_handler);
+
+        /**
+         * @brief Fetch track
+         *
+         * @param conn_id                   Connection ID to send subscribe
+         * @param track_handler             Track handler used for fetching
+         */
+        void FetchTrack(TransportConnId conn_id, std::shared_ptr<FetchTrackHandler> track_handler);
 
         /**
          * @brief Get the status of the Client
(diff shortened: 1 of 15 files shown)


include/quicr/detail/transport.h
```diff
@@ -268,6 +280,13 @@ namespace quicr {
                                 uint64_t track_alias,
                                 messages::SubscribeError error,
                                 const std::string& reason);
+        void SendFetch(ConnectionContext& conn_ctx,
```
TimEvens, line 283: Looks good, but this doesn't include start/end object.

include/quicr/fetch_track_handler.h
```diff
@@ -0,0 +1,164 @@
...
+        /**
+         * @brief  Status codes for the track
+         */
+        enum class Status : uint8_t
+        {
+            kOk = 0,
+            kNotConnected,
+            kFetchError,
+            kNotAuthorized,
+            kObjectNotAvailable,
```
TimEvens, line 36: Considering start/end object isn't part of send fetch, is this used for group not available?

TimEvens: How is it handled when some groups are available but the complete range is not? for example, fetch group 10 - 20, 10 - 15 are available, but 16 - 20 are not. Does the status change after receiving all of group 15?

I think what we want in this case is not a partial fetch.  I think if the fetch cannot be completely satisfied for the ranges specified, it should maybe return `'kGroupRangeNotAvailable` and `kObjectRangeNotAvailable`. In terms of MoQ, not really a good architecture because a suggested range that could be satisfied is not returned in the error response.

include/quicr/fetch_track_handler.h
```diff
@@ -0,0 +1,164 @@
...
+         * @brief Notification of received [full] data object
+         *
+         * @details Event notification to provide the caller the received full data object
+         *
+         * @warning This data will be invalided after return of this method
+         *
+         * @param object_headers    Object headers, must include group and object Ids
+         * @param data              Object payload data received, **MUST** match ObjectHeaders::payload_length.
+         */
+        virtual void ObjectReceived([[maybe_unused]] const ObjectHeaders& object_headers,
```
TimEvens, line 96: We need to figure out a way to detect if one of `ObjectReceived` or `PartialObjectReceived` is implemented. If one is implemented, then it's a valid fetch handler, if neither are implemented, it's an invalid handler and we shouldn't allow it.

include/quicr/fetch_track_handler.h
```diff
@@ -0,0 +1,164 @@
...
+         * @brief Set the status
+         * @param status Status of the subscribe
+         */
+        void SetStatus(Status status) noexcept
+        {
+            status_ = status;
+            StatusChanged(status);
+        }
+
+        Status status_{ Status::kObjectNotAvailable };
```
TimEvens, line 153: Default should probably be `kPendingFetchResponse` upon construction.  Transport will update it, but at least for the app, it'll know that it's still waiting on the transport to do something.

src/transport.cpp
```diff
@@ -369,12 +363,37 @@ namespace quicr {
...
+                              messages::GroupId end_group)
+    {
+        MoqFetch fetch;
+        fetch.subscribe_id = subscribe_id;
+        fetch.track_namespace = tfn.name_space;
+        fetch.track_name.assign(tfn.name.begin(), tfn.name.end());
+        fetch.priority = priority;
+        fetch.group_order = group_order;
+        fetch.start_group = start_group;
+        fetch.start_object = 0; // Request the start of the group
```
TimEvens, line 381: This works for our media use-case where media is kind of worthless without the start of group and the remaining objects in the group. In the case of other apps, they may want to fetch specific ranges of objects.

src/transport.cpp
```diff
@@ -369,12 +363,37 @@ namespace quicr {
...
+        fetch.end_group = end_group;
+        fetch.end_object = 0; // Request the whole group
+
+        Bytes buffer;
+        buffer.reserve(sizeof(MoqFetch) + tfn.name.size() + tfn.name_space.size());
```
TimEvens, line 386: We might want to revisit `sizeof(MoqFetch)` and how we determine the size of the struct.  Considering the values are put in the structure, we could have some size method in that structure that renders the 
size that will be needed for written bytes/buffer.

src/transport.cpp
```diff
@@ -369,12 +363,37 @@ namespace quicr {
...
+        buffer << fetch;
+
+        SendCtrlMsg(conn_ctx, buffer);
```
TimEvens, line 389: Should add a metric for number of fetch requests made.

src/transport.cpp
```diff
@@ -369,12 +363,37 @@ namespace quicr {
...
+
     void Transport::SendFetchError(ConnectionContext& conn_ctx,
                                    [[maybe_unused]] uint64_t subscribe_id,
                                    FetchError error,
                                    const std::string& reason)
     {
-
         auto fetch_err = MoqFetchError{};
         fetch_err.subscribe_id = 0x1;
         fetch_err.err_code = static_cast<uint64_t>(error);
```
TimEvens, line 399: Not part of our PR, but this should have had a metric as well for fetch failures.

TimEvens (approved):
Looks good, some minor comments.

include/quicr/fetch_track_handler.h
GhostofCookie: `¯\_(ツ)_/¯`

PR #331, opened 2024-11-20
Add smarter caching mechanism.

GhostofCookie:
Relates to https://github.com/Quicr/laps/issues/70


cmd/examples/server.cpp
```diff
@@ -539,59 +541,61 @@ class MyServer : public quicr::Server
...
+                          object.ttl,
+                          std::nullopt,
+                          object.extensions,
+                      };
+
+                      pub_track_h->PublishObject(obj_headers, object.data);
+                      std::this_thread::sleep_for(std::chrono::milliseconds(1));
+                  }
+              }
+          });
```
TimEvens, line 598: Not related to this change, but we need to clean up the bind publish track.  I have opened https://github.com/Quicr/libquicr/issues/332 to track that.

cmd/examples/server.cpp
```diff
@@ -614,7 +618,7 @@ class MyServer : public quicr::Server
         std::optional<quicr::Extensions> extensions;
         quicr::Bytes data;
     };
-    std::map<quicr::TrackNamespaceHash, std::multimap<quicr::messages::GroupId, CacheObject>> cache_;
+    std::map<quicr::TrackNamespaceHash, quicr::Cache<quicr::messages::GroupId, std::vector<CacheObject>>> cache_;
```
TimEvens, line 621: If we use set, it could ensure order based on object_id. Also, it would prevent duplicates, which is nice to dedup.

cmd/examples/server.cpp
GhostofCookie: Thanks for opening that!

cmd/examples/server.cpp
GhostofCookie: Ah, good idea, I'll add that! Thanks!

PR #346, opened 2024-12-16
Perf improvements

TimEvens:
Quick performance improvements to reduce cpu by half, resulting in almost double scale.

* Change transport ConnData to use shared pointer for data. This allows the pointer to be used all the way to the transport without having allocate/copy it around. 
* In time queue:
  *  Auto adjust the interval based on time duration to keep max buckets to 1000 or less. 
  * Change O(n) `Clear()` to clear only the buckets that are in use. When the queue is being serviced, the result should only be a few buckets being cleared instead of up to 1000 for every pop. 
* Add tick_service as a parameter to Server constructor to support reusing a single tick service instance.

### DIFF
--- a/include/quicr/detail/tick_service.h
+++ b/include/quicr/detail/tick_service.h
@@ -104,7 +104,7 @@ namespace quicr {
         std::atomic<bool> stop_{ false };
 
         /// Sleep delay in microseconds
-        const uint64_t sleep_delay_us_{ 10 };
+        const uint64_t sleep_delay_us_{ 333 };
 
         /// The thread to update ticks on.
         std::thread tick_thread_;
--- a/include/quicr/detail/time_queue.h
+++ b/include/quicr/detail/time_queue.h
@@ -249,9 +249,8 @@ namespace quicr {
         {
             queue_.clear();
 
-            //for (const auto& index: buckets_in_use_) {
-            for (auto& bucket : buckets_) {
-                bucket.clear();
+            for (const auto& index: buckets_in_use_) {
+                buckets_.at(index).clear();
             }
 
             buckets_in_use_.clear();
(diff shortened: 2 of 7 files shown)


GhostofCookie (approved):

include/quicr/detail/tick_service.h
```diff
@@ -104,7 +104,7 @@ namespace quicr {
         std::atomic<bool> stop_{ false };
 
         /// Sleep delay in microseconds
-        const uint64_t sleep_delay_us_{ 10 };
+        const uint64_t sleep_delay_us_{ 333 };
```
suhasHere, line 107: can we make it a config parameter ? 333ms might to be too slow

include/quicr/detail/time_queue.h
```diff
@@ -104,7 +107,7 @@ namespace quicr {
          */
         TimeQueue(size_t duration, size_t interval, std::shared_ptr<TickService> tick_service)
           : duration_{ duration }
-          , interval_{ interval }
+          , interval_{ (duration / interval > kMaxBuckets ? duration / 1000 : interval) }
```
suhasHere, line 110: duration/ kMaxBuckets

suhasHere (approved):
few minor nits. Otherwise LGTM

include/quicr/detail/tick_service.h
TimEvens: It's microseconds. Based on our usage now and longer term and considering that sleep relies on interrupts that will result in larger than sleep time variations, the change to 1/3 of a millisecond should be sufficient for our needs.  This provides granularity within ~2/3 of a millisecond.   

The original intent is that we would have 10 microsecond granularity, but that isn't working with sleep or select.  Later we'll revisit the tick service implementation to provide 10us granularity that hopefully avoids the interrupt/yield problem without incurring a busy loop that drives up CPU.

include/quicr/detail/time_queue.h
TimEvens: Thanks, updated.

PR #348, opened 2024-12-18
Add UnbindPublishTrack() to server API and fix memory leak issue

### LINKED ISSUE #332: BindPublishTrack() adds state that is not removed, need UnbindPublishTrack() method in server
Currently publish track handler is created to publish data to a subscriber. This track handler is not removed from state upon unsubscribe. 

A new method is needed to unbind the publish track handler and upon unsubscribe, it should be cleaned up.

TimEvens:
Resolves #332

### DIFF
--- a/src/server.cpp
+++ b/src/server.cpp
@@ -47,6 +47,38 @@ namespace quicr {
 
     void Server::ResolveSubscribe(ConnectionHandle, uint64_t, const SubscribeResponse&) {}
 
+    void Server::UnbindPublisherTrack(ConnectionHandle connection_handle,
+                                      const std::shared_ptr<PublishTrackHandler>& track_handler)
+    {
+        std::unique_lock lock(state_mutex_);
+
+        auto conn_it = connections_.find(connection_handle);
+        if (conn_it != connections_.end()) {
+            auto th = TrackHash(track_handler->GetFullTrackName());
+            SPDLOG_LOGGER_DEBUG(
+              logger_,
+              "Server publish track conn_id: {0} full_name_hash: {} namespace_hash: {} name_hash: {} unbind",
+              connection_handle,
+              th.track_fullname_hash,
+              th.track_namespace_hash,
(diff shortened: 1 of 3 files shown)


src/server.cpp
```diff
@@ -47,6 +47,38 @@ namespace quicr {
 
     void Server::ResolveSubscribe(ConnectionHandle, uint64_t, const SubscribeResponse&) {}
 
+    void Server::UnbindPublisherTrack(ConnectionHandle connection_handle,
+                                      const std::shared_ptr<PublishTrackHandler>& track_handler)
+    {
+        std::unique_lock lock(state_mutex_);
+
+        auto conn_it = connections_.find(connection_handle);
+        if (conn_it != connections_.end()) {
```
GhostofCookie, line 56: I would maybe flip the condition and return early to avoid indenting the entire function, but that's just a nit.

src/server.cpp
```diff
@@ -82,8 +114,9 @@ namespace quicr {
                                              false);
 
         // Setup the function for the track handler to use to send objects with thread safety
+        std::weak_ptr weak_track_handler(track_handler);
         track_handler->publish_object_func_ =
-          [&, track_handler, subscribe_id = track_handler->GetSubscribeId(), cb = std::move(callback)](
+          [&, weak_track_handler, subscribe_id = track_handler->GetSubscribeId(), cb = std::move(callback)](
```
GhostofCookie, line 119: Good catch! Subscribe id isn't used here either, should we remove it?

GhostofCookie (approved):

src/server.cpp
TimEvens: makes sense, updated.

src/server.cpp
TimEvens: Looks like it wasn't an issue to keep it, so I removed it.

GhostofCookie (approved):

PR #349, opened 2025-01-06
Remove boringSSL, use platform/os ssl instead

TimEvens:
This removes boringSSL and uses cmake findSSL to find the package. This will use static or dynamic linking for crypto. This PR allows others to build using their provided SSL package.


GhostofCookie (approved):
Without building it myself this looks good!

PR #350, opened 2025-01-08
Add Shared Memory for incoming data.

### LINKED ISSUE #351: re-usable shared memory access (aka reducing number of copies)

GhostofCookie:
Closes #351


test/shared_memory.cpp
```diff
@@ -0,0 +1,38 @@
+// SPDX-FileCopyrightText: Copyright (c) 2024 Cisco Systems
+// SPDX-License-Identifier: BSD-2-Clause
+
+#include <doctest/doctest.h>
+
+#include "quicr/shared_memory.h"
+
```
TimEvens, line 7: Maybe a few more tests.  I was using the below to test this, which is working well. 

```c++
TEST_CASE("SharedMemory Multiples")
{
    auto buffer = quicr::SharedMemory::Create();

    std::string s1 = "one";
    std::string s2 = " two";
    std::string s3 = " three";

    buffer->Push({reinterpret_cast<uint8_t*>(&s1[0]), s1.size()});
    buffer->Push({reinterpret_cast<uint8_t*>(&s2[0]), s2.size()});
    buffer->Push({reinterpret_cast<uint8_t*>(&s3[0]), s3.size()});

    std::vector<uint8_t> v;
    for (auto value : *buffer)
    {
        v.push_back(value);
    }

    CHECK_EQ(v.size(), s1.size() + s2.size() + s3.size());

    CHECK_EQ(v.at(5), 'w');
}
```

TimEvens (commented):
This is great. Although, we will likely not have more than one slice. It's nice to have this though, because we have that crazy situation that QUIC/MoQ causes where we may not have enough data to parse the needed track alias out to find the subscribe handler. 

This is great as the shared memory, but we need something very similar for `PayloadData` that the subscribe handler will use.  In other words, subscribe handler will need to `Push()` shared pointers as slices.  `PayloadData` also needs to have a method to set the begin iterator and end iterator, which happens after construct.

TimEvens (approved):

PR #352, opened 2025-01-20
Fixup doc warnings

RichLogan:
Fix up some minor doc related warnings that can show up in build output.

### DIFF
--- a/include/quicr/hash.h
+++ b/include/quicr/hash.h
@@ -73,6 +73,7 @@ namespace quicr {
      *
      * @returns The hash of the given string.
      */
+    [[maybe_unused]]
     static std::uint64_t hash(const std::string_view& str)
     {
         return hash(std::vector<uint8_t>{ str.begin(), str.end() });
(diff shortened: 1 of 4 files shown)


include/quicr/hash.h
```diff
@@ -73,6 +73,7 @@ namespace quicr {
      *
      * @returns The hash of the given string.
      */
+    [[maybe_unused]]
     static std::uint64_t hash(const std::string_view& str)
```
GhostofCookie, line 77: May be better to just remove it instead.

GhostofCookie (approved):

include/quicr/hash.h
TimEvens: I have to check, but laps might be using this one.  Maybe add a test?.?   Seems like we don't have specific tests for hash, but do have some uses in track_namespace tests.

GhostofCookie: I see the warning when I build laps, so I don't think it is being used.

RichLogan: I can't see anywhere that uses it, do we vote to remove? Can do it in a subsequent PR.

PR #442, opened 2025-02-18
Adding latest group and object info to send sub ok.

### LINKED ISSUE #361: Subscribe OK to provide state of the group

GhostofCookie:
Closes #361

### DIFF
--- a/src/server.cpp
+++ b/src/server.cpp
@@ -60,7 +60,7 @@ namespace quicr {
                   conn_it->second,
                   subscribe_id,
                   kSubscribeExpires,
-                  false,
+                  subscribe_response.largest_group.has_value() && subscribe_response.largest_object.has_value(),
                   subscribe_response.largest_group.has_value() ? subscribe_response.largest_group.value() : 0,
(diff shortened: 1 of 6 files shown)


TimEvens (approved):
Looks good, but it would be nice if the subscribe ok debug logged the group/object id. This way we can see it in qclient.

src/server.cpp
```diff
@@ -56,7 +56,13 @@ namespace quicr {
 
         switch (subscribe_response.reason_code) {
             case SubscribeResponse::ReasonCode::kOk: {
-                SendSubscribeOk(conn_it->second, subscribe_id, kSubscribeExpires, false);
+                SendSubscribeOk(
+                  conn_it->second,
+                  subscribe_id,
+                  kSubscribeExpires,
+                  false,
```
suhasHere, line 63: this has to be set based on latest group and latest object being present or not

suhasHere (changes requested):

suhasHere (approved):

PR #443, opened 2025-02-18
Add New Group message

### LINKED ISSUE #406: active speaker switching flow - new group - lib quicr implementation
due 2/20

GhostofCookie:
Closes #406

### DIFF
--- a/include/quicr/detail/messages.h
+++ b/include/quicr/detail/messages.h
@@ -67,6 +67,8 @@ namespace quicr::messages {
 
         CLIENT_SETUP = 0x40,
         SERVER_SETUP,
+
+        kNewGroup,
     };
 
     enum class DataMessageType : uint8_t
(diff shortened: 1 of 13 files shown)


include/quicr/detail/messages.h
```diff
@@ -172,6 +174,19 @@ namespace quicr::messages {
...
     Bytes& operator<<(Bytes& buffer, const MoqServerSetup& msg);
 
+    //
+    // New Group
+    //
+
+    struct NewGroupRequest
+    {
+        uint64_t subscribe_id;
+        uint64_t track_alias;
```
suhasHere, line 184: we don't need track alias here

TimEvens: The reason is that the client must have the track alias to match received objects. The client maintains the track alias and can provide it. The server has the track alias as well, but forwarding of new group goes the the origin publisher which may not even be on the same relay. Forwarding to the origin publisher involves looking up the track alias. **The track alias is required**.  

That being said, we can of course take a performance hit and do a double lookup for every new group request. The double lookup is to lookup the subscribe ID of the incoming request to find the track alias that the subscribe ID is associated to. Then to do what we would have done had the original request had track alias.

TimEvens (approved):
LGTM considering seeing it in action.

include/quicr/subscribe_track_handler.h
```diff
@@ -224,6 +234,8 @@ namespace quicr {
         messages::FilterType filter_type_;
         StreamBuffer<uint8_t> stream_buffer_;
         uint64_t current_stream_id_{ 0 };
+        messages::GroupId latest_group_id_;
+        messages::ObjectId latest_object_id_;
```
suhasHere, line 238: can we make them std::optional

PR #444, opened 2025-02-19
Even header extension support

RichLogan:
This is a quick addition for interop of the "short" header extensions that consist of a single varint value when the extension type is even (`%2=0`). I would suggest we subsequently add a more advanced data structure for the `quicr::Extensions` type that can actually check the validity of extensions given these rules. Right now, it's still just a map of bytes and so the functions that accept it are naive to this change until they are put on/taken off the wire. 

I added a `ReadUintV` function to return the *encoded* varint for calls who want to inspect it. If that's not a good idea, no big deal, I'll change it back and construct in place.

### DIFF
--- a/src/messages.cpp
+++ b/src/messages.cpp
@@ -55,12 +55,24 @@ namespace quicr::messages {
                 }
                 current_tag = tag;
             }
-            auto val = buffer.DecodeBytes();
-            if (!val) {
-                count -= completed;
-                return false;
+
+            if (current_tag.value() % 2 == 0) {
+                uint64_t value{ 0 };
+                if (!ParseUintVField(buffer, value)) {
+                    count -= completed;
+                    return false;
+                }
+                std::vector<std::uint8_t> bytes(sizeof(std::uint64_t));
+                std::memcpy(bytes.data(), &value, sizeof(std::uint64_t));
+                extensions.value()[current_tag.value()] = std::move(bytes);
+            } else {
+                auto val = buffer.DecodeBytes();
+                if (!val) {
+                    count -= completed;
+                    return false;
+                }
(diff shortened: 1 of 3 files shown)


src/messages.cpp
```diff
@@ -55,12 +55,26 @@ namespace quicr::messages {
...
+            if (current_tag.value() % 2 == 0) {
+                auto val = buffer.ReadUintV();
+                if (!val) {
+                    count -= completed;
+                    return false;
+                }
+                const std::size_t size = val.value().size();
+                std::vector<std::uint8_t> bytes(size);
+                auto decoded = uint64_t(val.value());
+                std::memcpy(bytes.data(), &decoded, size);
```
GhostofCookie, line 68:
```suggestion
                std::vector<std::uint8_t> bytes;
                bytes.assign(val->begin(), val->end());
```

src/messages.cpp
```diff
@@ -89,6 +103,14 @@ namespace quicr::messages {
         buffer << UintVar(extensions.value().size());
         for (const auto& extension : extensions.value()) {
             buffer << UintVar(extension.first);
+            if (extension.first % 2 == 0) {
+                // Even types are a single varint value.
+                std::uint64_t val = 0;
+                std::memcpy(&val, extension.second.data(), std::min(extension.second.size(), sizeof(std::uint64_t)));
```
GhostofCookie, line 109: Can size ever be larger than 8?

RichLogan: Not according to the spec, but it's not enforced on our side when you pass something in.

GhostofCookie (approved):

suhasHere (approved):

PR #453, opened 2025-02-19
Conform messages to style guide.

GhostofCookie:


TimEvens (approved):
LGTM, I believe there might be a few updates needed in laps, but that shouldn't be a problem.

PR #454, opened 2025-02-19
Adds qserver support for prefix matching

### LINKED ISSUE #378: Server supports prefix matching on namespaces
2/21

TimEvens:
Closes #378

### DIFF
--- a/cmd/examples/server.cpp
+++ b/cmd/examples/server.cpp
@@ -384,28 +384,31 @@ class MyServer : public quicr::Server
 
         // Check if there are any subscribes. If so, send subscribe to announce for all tracks matching namespace
         for (const auto& [ns, sub_tracks]: qserver_vars::subscribe_active) {
-            if (ns.HasSamePrefix(track_namespace)) {
-                for (const auto& [track_name, si] : sub_tracks) {
-                    if (si.size()) { // Have subscribes
-                        auto& a_si = *si.begin();
-                        if (anno_tracks.find(a_si.track_alias) == anno_tracks.end()) {
-                            SPDLOG_INFO("Sending subscribe to announcer connection handle: {0} subscribe track_alias: {1}",
-                                        connection_handle,
-                                        a_si.track_alias);
+            if (!ns.HasSamePrefix(track_namespace)) {
+                continue;
+            }
 
-                            anno_tracks.insert(a_si.track_alias); // Add track to state
+            for (const auto& [track_name, si] : sub_tracks) {
+                if (si.size()) { // Have subscribes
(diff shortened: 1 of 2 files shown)


cmd/examples/server.cpp
```diff
@@ -383,28 +380,29 @@ class MyServer : public quicr::Server
...
-
-                        const auto pub_track_h = qserver_vars::subscribes[a_who.track_alias][a_who.connection_handle];
-
-                        auto sub_track_handler =
-                          std::make_shared<MySubscribeTrackHandler>(pub_track_h->GetFullTrackName());
-
-                        SubscribeTrack(connection_handle, sub_track_handler);
-                        qserver_vars::pub_subscribes[a_who.track_alias][connection_handle] = sub_track_handler;
+        for (const auto& [ns, sub_tracks]: qserver_vars::subscribe_active) {
+            if (ns.HasSamePrefix(track_namespace)) {
```
GhostofCookie, line 387: Recommend flipping the condition and continuing to avoid indenting the whole block even more.

cmd/examples/server.cpp
```diff
@@ -469,7 +467,8 @@ class MyServer : public quicr::Server
             return;
         }
 
-        auto th = quicr::TrackHash(track_h->GetFullTrackName());
+        const auto& tfn = track_h->GetFullTrackName();
+        auto th = quicr::TrackHash(tfn);
```
GhostofCookie, line 471: Is this used? I don't see it.

cmd/examples/server.cpp
```diff
@@ -570,58 +569,64 @@ class MyServer : public quicr::Server
...
-                  std::chrono::duration_cast<std::chrono::milliseconds>(now - last_subscription_time_.value()).count();
-                if (elapsed > kSubscriptionDampenDurationMs_) {
-                    // send subscription update
-                    auto& sub_track_h = qserver_vars::pub_subscribes[th.track_fullname_hash][conn_h];
-                    if (sub_track_h == nullptr) {
-
-                        return;
+        bool success = false;
+        for (auto& [ns, conns]: qserver_vars::announce_active) {
+            if (ns.HasSamePrefix(track_full_name.name_space)) {
```
GhostofCookie, line 581: Same as before, flip the condition and continue and avoid indenting the whole thing.

cmd/examples/server.cpp
TimEvens: Yeah, it's used via lines starting at 483.  Namespace is used now instead of namespace hash, which was part of th.

GhostofCookie (approved):

PR #467, opened 2025-02-21
Adding fetch cancel.

### LINKED ISSUE #450: Fetch cancel
Implement unbind publish track and correctly support fetch cancel (terminate the active thread).

GhostofCookie:
Closes #450

### DIFF
--- a/cmd/examples/server.cpp
+++ b/cmd/examples/server.cpp
@@ -731,7 +731,7 @@ class MyServer : public quicr::Server
                       SPDLOG_INFO("Fetching group: {} object: {}", object.headers.group_id, object.headers.object_id);
 
                       pub_track_h->PublishObject(object.headers, object.data);
-                      std::this_thread::sleep_for(std::chrono::milliseconds(1000));
+                      std::this_thread::sleep_for(std::chrono::milliseconds(1));
                   }
               }
           });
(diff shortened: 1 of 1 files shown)


cmd/examples/server.cpp
```diff
@@ -121,6 +122,10 @@ namespace qserver_vars {
      */
     std::shared_ptr<quicr::ThreadedTickService> tick_service = std::make_shared<quicr::ThreadedTickService>();
 
+    /**
+     * @brief Map of atomic bools to mark if a fetch thread should be interrupted.
+     */
+    std::unordered_map<uint64_t, std::atomic_bool> stop_fetch;
```
TimEvens, line 128: Will need to add back unordered_map header. Recent change removes it because map is used instead. Is there a perf gain here using unordered or could this be map?

cmd/examples/server.cpp
```diff
@@ -705,15 +710,26 @@ class MyServer : public quicr::Server
 
         const auto th = quicr::TrackHash(track_full_name);
 
+        qserver_vars::stop_fetch.emplace(std::make_pair(subscribe_id, false));
+
         std::thread retrieve_cache_thread(
           [=, cache_entries = qserver_vars::cache.at(th.track_fullname_hash).Get(attrs.start_group, attrs.end_group)] {
+              defer(UnbindPublisherTrack(connection_handle, pub_track_h));
```
TimEvens, line 717: Nice use of defer.

cmd/examples/server.cpp
GhostofCookie: The difference would be unnoticeable.

PR #472, opened 2025-02-21
Fix object datagram encode/decode order.

### LINKED ISSUE #465: OBJECT_DATAGRAM - out of ordered encoding/decoding
|MESSAGE|NOTES|
|---|---|
|OBJECT_DATAGRAM | seems like we are doing this out of order for object status and extension headers |

Most likely difference between v7 and v8.

Draft v8:

```
   OBJECT_DATAGRAM {
     Track Alias (i),
     Group ID (i),
     Object ID (i),
     Publisher Priority (8),
     Extension Count (i),
     [Extension headers (...)],
     Object Payload Length (i),
     [Object Status (i)],
     Object Payload (..),
   }
```

Code:

```C++
    Bytes& operator<<(Bytes& buffer, const MoqObjectDatagram& msg)
    {
        buffer << UintVar(static_cast<uint64_t>(DataMessageType::OBJECT_DATAGRAM));
        buffer << UintVar(msg.track_alias);
        buffer << UintVar(msg.group_id);
        buffer << UintVar(msg.object_id);
        buffer.push_back(msg.priority);
        if (msg.payload.empty()) {
            // empty payload needs a object status to be set
            buffer << UintVar(static_cast<uint8_t>(msg.object_status));
        } else {
            buffer << msg.payload;
        }
        buffer << msg.payload;
        return buffer;
    }
```

Need to move payload length after extensions and make sure `PushExtensions()` includes a count, even if zero.

GhostofCookie:
Closes #465

### DIFF
--- a/src/messages.cpp
+++ b/src/messages.cpp
@@ -914,12 +914,11 @@ namespace quicr::messages {
 
         buffer << UintVar(msg.payload.size());
         if (msg.payload.empty()) {
-            // empty payload needs a object status to be set
             buffer << UintVar(static_cast<uint8_t>(msg.object_status));
-        } else {
-            buffer << msg.payload;
         }
+
         buffer << msg.payload;
+
         return buffer;
     }
 
--- a/src/messages.cpp
+++ b/src/messages.cpp
@@ -910,13 +910,13 @@ namespace quicr::messages {
         buffer << UintVar(msg.group_id);
         buffer << UintVar(msg.object_id);
         buffer.push_back(msg.priority);
+        PushExtensions(buffer, msg.extensions);
+
         buffer << UintVar(msg.payload.size());
         if (msg.payload.empty()) {
             // empty payload needs a object status to be set
             buffer << UintVar(static_cast<uint8_t>(msg.object_status));
-            PushExtensions(buffer, msg.extensions);
         } else {
-            PushExtensions(buffer, msg.extensions);
             buffer << msg.payload;
         }
         buffer << msg.payload;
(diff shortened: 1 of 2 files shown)


src/messages.cpp
```diff
@@ -910,13 +910,13 @@ namespace quicr::messages {
...
         buffer << UintVar(msg.payload.size());
         if (msg.payload.empty()) {
             // empty payload needs a object status to be set
             buffer << UintVar(static_cast<uint8_t>(msg.object_status));
-            PushExtensions(buffer, msg.extensions);
         } else {
-            PushExtensions(buffer, msg.extensions);
             buffer << msg.payload;
         }
         buffer << msg.payload;
```
TimEvens, line 922: Why are we adding twice `msg.payload`?

src/messages.cpp
```diff
@@ -910,16 +910,15 @@ namespace quicr::messages {
...
         if (msg.payload.empty()) {
-            // empty payload needs a object status to be set
             buffer << UintVar(static_cast<uint8_t>(msg.object_status));
-            PushExtensions(buffer, msg.extensions);
-        } else {
-            PushExtensions(buffer, msg.extensions);
-            buffer << msg.payload;
         }
+
         buffer << msg.payload;
```
TimEvens, line 920: I suppose it's fine to call to encode payload when it's zero in size but we could have returned from the payload empty check block to prevent an attempt on zero size payload.

TimEvens (approved):

PR #473, opened 2025-02-21
Encode announce params.

### LINKED ISSUE #464: ANNOUNCE - not encoding parameters
|MESSAGE|NOTES|
|---|---|
| ANNOUNCE | we are not encoding parameters, we do decode parameters |

C++ not encoding parameters. Although they are decoded.

```C++

    Bytes& operator<<(Bytes& buffer, const MoqAnnounce& msg)
    {
        Bytes payload;
        payload << msg.track_namespace;
        payload << UintVar(static_cast<uint64_t>(0));

        buffer << UintVar(static_cast<uint64_t>(ControlMessageType::ANNOUNCE));
        buffer << UintVar(payload.size());
        buffer << payload;

        return buffer;
    }
```

GhostofCookie:
Closes #464

### DIFF
--- a/src/messages.cpp
+++ b/src/messages.cpp
@@ -639,7 +639,6 @@ namespace quicr::messages {
     {
         Bytes payload;
         payload << msg.track_namespace;
-        payload << UintVar(static_cast<uint64_t>(0));
 
         payload << UintVar(msg.params.size());
         for (const auto& param : msg.params) {
--- a/src/messages.cpp
+++ b/src/messages.cpp
@@ -641,6 +641,11 @@ namespace quicr::messages {
         payload << msg.track_namespace;
         payload << UintVar(static_cast<uint64_t>(0));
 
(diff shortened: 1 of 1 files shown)


src/messages.cpp
```diff
@@ -641,6 +641,11 @@ namespace quicr::messages {
         payload << msg.track_namespace;
         payload << UintVar(static_cast<uint64_t>(0));
```
TimEvens, line 642: This should be removed since line #644 adds this again.

src/messages.cpp
```diff
@@ -641,6 +641,11 @@ namespace quicr::messages {
...
 
+        payload << UintVar(msg.params.size());
```
TimEvens, line 644: This one replaces line #642

TimEvens (changes requested):

TimEvens (approved):

PR #475, opened 2025-02-21
Remove unused path parameter.

### LINKED ISSUE #457: CLIENT_SETUP and SERVER_SETUP missing "path_parameter"
Message parameter missing in CLIENT_SETUP and SERVER_SETUP.

|MESSAGE | NOTES |
| --- | --- |
|CLIENT_SETUP | "Setup Parameter" "path_parameter" that doesn't seem to be used |
| SERVER_SETUP | "Setup Parameter" "path_parameter" that doesn't seem to be used |

Not sure if it is a big deal but both of these messages are defined in code to have a "path_parameter":

```C++
    struct MoqClientSetup
    {
        uint64_t num_versions{ 0 };
        std::vector<Version> supported_versions;
        MoqParameter role_parameter;
        MoqParameter path_parameter;
        MoqParameter endpoint_id_parameter;
    };

    struct MoqServerSetup
    {
        Version selection_version;
        MoqParameter role_parameter;
        MoqParameter path_parameter;
        MoqParameter endpoint_id_parameter;
    };
```

The encoding for these messages does not encode the "path_parameter" but the decoding does decode this parameter.

GhostofCookie:
Closes #457

### DIFF
--- a/include/quicr/detail/messages.h
+++ b/include/quicr/detail/messages.h
@@ -154,7 +154,6 @@ namespace quicr::messages {
         uint64_t num_versions{ 0 };
         std::vector<Version> supported_versions;
         Parameter role_parameter;
-        Parameter path_parameter;
         Parameter endpoint_id_parameter;
     };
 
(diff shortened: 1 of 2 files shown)


TimEvens (approved):
LGTM, we'll treat at unknown/ignored param right now. When/if we actually use it, we can add it back.

PR #518, opened 2025-03-06
Handle invalid message

TimEvens:
Fixes crashes due to invalid message encoding

### DIFF
--- a/src/messages.cpp
+++ b/src/messages.cpp
@@ -126,7 +126,7 @@ namespace quicr::messages {
         uint64_t size = 0;
         buffer = buffer >> size;
         if (size > buffer.size()) {
-            throw std::invalid_argument("Size of vector is larger than buffer size");
+            throw std::runtime_error("Namespace entry size is larger than buffer size");
         }
         value.assign(buffer.begin(), std::next(buffer.begin(), size));
         return buffer.subspan(value.size());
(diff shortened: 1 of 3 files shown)


RichLogan (approved):

src/messages.cpp
```diff
@@ -125,6 +125,9 @@ namespace quicr::messages {
     {
         uint64_t size = 0;
         buffer = buffer >> size;
+        if (size > buffer.size()) {
+            throw std::runtime_error("Namespace entry size is larger than buffer size");
```
GhostofCookie, line 129: You could throw std::out_of_range for extra clarity.

src/messages.cpp
```diff
@@ -426,6 +429,10 @@ namespace quicr::messages {
 
     BytesSpan operator>>(BytesSpan buffer, Subscribe& msg)
     {
+        if (buffer.size() < 1) {
+            throw std::runtime_error("Invalid subscribe data to decode, too small");
```
GhostofCookie, line 433: Same here.

GhostofCookie (approved):
Comments optional, looks great.

src/messages.cpp
TimEvens: updated.

src/messages.cpp
TimEvens: updated.

PR #519, opened 2025-03-12
Add FetchAvailability result

RichLogan:
This is 50% a proposal and 50% a PR. 

FETCH_OK requires some attributes that an implementation needs to provide:

> End Of Track: 1 if all objects have been published on this track, so the Largest Group ID and Object Id indicate the last Object in the track, 0 if not.
> Largest Group ID: The largest Group ID available for this track.
> Largest Object ID: The largest Object ID available within the largest Group ID for this track.

This allows the FetchReceived implementor to provide them, if the fetch can be resolved. 

A similar API will be needed for JoiningFetch (for matching subscription lookup that needs to be implementor provided), so wanted to float this separately.

### DIFF
--- a/cmd/examples/server.cpp
+++ b/cmd/examples/server.cpp
@@ -770,10 +770,10 @@ class MyServer : public quicr::Server
      *
      * @returns true if the range of groups and objects exist in the cache, otherwise returns false.
      */
-    bool FetchReceived([[maybe_unused]] quicr::ConnectionHandle connection_handle,
-                       [[maybe_unused]] uint64_t subscribe_id,
-                       const quicr::FullTrackName& track_full_name,
-                       const quicr::FetchAttributes& attrs) override
+    std::optional<FetchAvailability> FetchReceived([[maybe_unused]] quicr::ConnectionHandle connection_handle,
+                                                   [[maybe_unused]] uint64_t subscribe_id,
+                                                   const quicr::FullTrackName& track_full_name,
+                                                   const quicr::FetchAttributes& attrs) override
     {
         SPDLOG_INFO("Received Fetch for conn_id: {} subscribe_id: {} start_group: {} end_group: {}",
                     connection_handle,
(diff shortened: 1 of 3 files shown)


src/server.cpp
```diff
@@ -548,7 +551,8 @@ namespace quicr {
                     .end_object = msg.end_object,
                 };
 
-                if (!FetchReceived(conn_ctx.connection_handle, msg.subscribe_id, tfn, attrs)) {
+                const auto availability = FetchReceived(conn_ctx.connection_handle, msg.subscribe_id, tfn, attrs);
+                if (!availability) {
```
GhostofCookie, line 555: Slight nit, but `available` would read better.

cmd/examples/server.cpp
```diff
@@ -795,13 +795,21 @@ class MyServer : public quicr::Server
...
-        return std::any_of(groups.begin(), groups.end(), [&](const auto& group) {
+        const bool available = std::any_of(groups.begin(), groups.end(), [&](const auto& group) {
             return !group->empty() && group->begin()->headers.object_id <= attrs.start_object &&
                    std::prev(group->end())->headers.object_id >= (attrs.end_object - 1);
         });
+        if (!available) {
+            SPDLOG_WARN("No objects found for requested range");
+            return std::nullopt;
+        }
+        const auto last_cached = groups.back()->end()->headers;
```
GhostofCookie, line 809: This will crash, since `end()` is one past the actual end.
```suggestion
        const auto last_cached = std::prev(groups.back()->end())->headers;
```

RichLogan: lol I made the exact same mistake on `groups`, fixed it, and then didn't change the one literally right next to it.

RichLogan:
Given what was discussed today, will change this to be a more generic cache lookup callback.

RichLogan:
Subsumed by #527

PR #524, opened 2025-03-13
Fetch range to match spec

RichLogan:
Update FETCH range calculation to match the draft. Here is the corresponding text. 

Wording:

> Fetch specifies an inclusive range of Objects starting at StartObject in StartGroup and ending at EndObject in EndGroup. 

Message definitions:

> StartGroup: The start Group ID.
> StartObject: The start Object ID.
> EndGroup: The end Group ID.
> EndObject: The end Object ID, plus 1. A value of 0 means the entire group is requested.

I have made an assumption that when it says "inclusive...end object" in the former, it is referring to after the resolution of the `-1` or the `0 == whole group` from what's in the actual message. I think this text is slightly confusing because they use the `EndObject` name in the text that matches the field defined below, rather than the words `end Object ID` as used in the expanded message text. Maybe I will raise that as an issue. 

The changes I've made:

1. To make things really clear, fetch attributes is always inclusive. If end object isn't set, it's the whole group.
2. Updated the qserver calculations to reflect that it is an inclusive range.
3. I left the cache lookup as exclusive end, to match STL, and so added 1 to the end group value.
4. I added an exception if you request an invalid range from the cache, because we were silently returning null if you asked for an intra range previously i.e if the FETCH asked for 0:0 -> 0:5 we would ask the cache for group 0 -> group 0 and it would return null because start == end. That was a bug that wasn't immediately obvious because of the silent fail. 
5. I added tests for the cache while I was validating what was going on, and figured I'd leave them there. 
6. Changed the tick service slightly so I could mock it in the test.

I have validated that the correct things are retrieved from the cache now with qclient and qserver. I.E if all objects were `[(0,0), (0,1), (1,0), (1,1)]` then a FETCH **message** of Start Group: 0, Start Object: 0, End Group: 1, End Object: 1 would return `[(0,0), (0,1), (1,0)]` which I believe  to be correct given my interpretation of the draft.


include/quicr/common.h
```diff
@@ -77,12 +77,13 @@ namespace quicr {
...
-        uint64_t start_object;            ///< Fetch starting object in group
-        uint64_t end_group;               ///< Fetch final group in range
-        uint64_t end_object;              ///< Fetch final object in group
+        uint8_t priority;                 ///< Priority
+        messages::GroupOrder group_order; ///< Group order
+        uint64_t start_group;             ///< Starting group to fetch from, inclusive.
+        uint64_t start_object;            ///< Starting object in start_group to fetch from, inclusive.
+        uint64_t end_group;               ///< Final group to fetch from, inclusive.
+        std::optional<std::uint64_t>
+          end_object; ///< Final object in end_group to fetch from, inclusive. If null, fetch all objects in end_group.
```
GhostofCookie, line 86: I'll just state it again here, not sure it needs optional
```suggestion
        std::uint64_t end_object = -1; ///< Final object in end_group to fetch from, inclusive. If invalid, fetch all objects in end_group.
```

Also, not sure I agree with the comment. If this is null? I thought the spec said if group was 0

test/cache.cpp
```diff
@@ -0,0 +1,59 @@
...
+        REQUIRE(retrieved.size() == 1);
+        CHECK(*retrieved[0] == expected);
+
+        // Get both keys.
+        retrieved = cache.Get(target_key, target_key + 2);
+        REQUIRE(retrieved.size() == 2);
+        CHECK(*retrieved[0] == expected);
+        CHECK(*retrieved[1] == expected_second);
+    }
+}
```
GhostofCookie, line 59: No newline? Naughty naughty....

GhostofCookie (approved):
The changes seem sensible. My opinion on the optional is itself, optional.

include/quicr/common.h
RichLogan: I agree it's not great. If this was something like Swift, something like this might get the intent across better than `?` or a `-1`.

```swift
enum EndFilter { case entireGroup, case endObject(messages::ObjectId) }
EndFilter end_filter;
```

Just for posterity, this value is specifically different to the value that is in the wire message, this is after resolving that back to the inclusive range that the spec talks about. A wire message with an end object of 1 means you want your fetch range (inclusive) to end with and include the 0th object. In that case, end object here *would* be 0, reflecting the inclusive range.

I was trying to make a distinction here between the wire format optimisation of the 0 == whole group vs. endObject -= 1 and the inclusive range the spec refers to.

cmd/examples/server.cpp
```diff
@@ -839,9 +841,9 @@ class MyServer : public quicr::Server
                           return;
                       }
 
-                      if ((object.headers.group_id < attrs.start_group || object.headers.group_id >= attrs.end_group) ||
+                      if ((object.headers.group_id < attrs.start_group || object.headers.group_id > attrs.end_group) ||
                           (object.headers.object_id < attrs.start_object ||
-                           object.headers.object_id >= attrs.end_object))
+                           (attrs.end_object.has_value() && object.headers.object_id > attrs.end_object.value())))
```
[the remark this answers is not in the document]
RichLogan: If storage or performance is a concern then I'm 100% down to give way but I picked an optional because I preferred the semantic meaning of "no end object means no end object"  vs `-1 / ::max() means no end object`.

cmd/examples/server.cpp
```diff
@@ -800,7 +800,8 @@ class MyServer : public quicr::Server
 
         return std::any_of(groups.begin(), groups.end(), [&](const auto& group) {
             return !group->empty() && group->begin()->headers.object_id <= attrs.start_object &&
-                   std::prev(group->end())->headers.object_id >= (attrs.end_object - 1);
+                   (!attrs.end_object.has_value() ||
+                    std::prev(group->end())->headers.object_id >= attrs.end_object.value());
```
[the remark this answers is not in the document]
RichLogan: Ah, this is because it's a `set`, it has no `back()`. The vector of sets does though.

RichLogan:
I'm sticking with the optional, because while -1 / max() works out of the box for the checking what's valid test, it doesn't work out of the box for the availability check when you're looking to see if you have >= end_object in the cache. At that point, you would have to handle `end_object == ::max()` as a specific value signalling everything, which semantically feels worse than an optional. That being said, if performance/overhead is a concern, 100% happy to do exactly that.

PR #525, opened 2025-03-13
Clear TX object on stream change and ensure start of stream has enough data to start with a few uintvars

TimEvens:


src/transport_picoquic.cpp
```diff
@@ -1171,6 +1175,10 @@ PicoQuicTransport::SendStreamBytes(DataContext* data_ctx, uint8_t* bytes_ctx, si
         return;
     }
 
+    if (max_len < 20 && data_ctx == nullptr && data_ctx->tx_start_stream) {
```
GhostofCookie, line 1178: Magic 20?

GhostofCookie (approved):

PR #529, opened 2025-03-21
ESP Compatibility

GhostofCookie:


CMakeLists.txt
```diff
@@ -53,6 +53,12 @@ endif()
 ###
 set (BUILD_SHARED_LIBS OFF)
 set (BUILD_STATIC_LIBS ON)
+
+if (NOT PLATFORM_ESP_IDF)
+    set (CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
+    list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/dependencies/transport/cmake)
+endif()
+
 set (CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
```
TimEvens, line 62: Looks like we are setting this twice still.

dependencies/CMakeLists.txt
```diff
@@ -32,47 +32,55 @@ if(BUILD_TESTING AND QUICR_BUILD_TESTS AND NOT TARGET doctest)
...
+            # Continue with mbedtls inclusion.
+            option(ENABLE_TESTING OFF)
+            option(ENABLE_PROGRAMS OFF)
+            set(MBEDTLS_AS_SUBPROJECT ON)
+            add_subdirectory(mbedtls)
+
+            set(MBEDTLS_ROOT_DIR ${CMAKE_CURRENT_SOURCE_DIR}/mbedtls)
+            set(MBEDTLS_INCLUDE_DIRS ${CMAKE_CURRENT_SOURCE_DIR}/mbedtls ${CMAKE_CURRENT_SOURCE_DIR}/mbedtls/include)
+
+        elseif(PLATFORM_ESP_IDF)
```
TimEvens, line 79: Seems like IDF_PATH is required and will be set if building for ESP_IDF.  Wonder, do we need `PLATFORM_ESP_IDF` when we could check if IDF_PATH is set?

dependencies/CMakeLists.txt
```diff
@@ -90,16 +98,26 @@ if (NOT PLATFORM_ESP_IDF)
...
-        set(OPENSSL_INCLUDE_DIR "") # Picoquic will still error if this is not set
-        set(PICOQUIC_PTLS_SUBMODULE ON)
-        option(WITH_OPENSSL OFF)
-        set(picoquic_BUILD_TESTS OFF)
-        add_subdirectory(picoquic)
-        add_dependencies(picoquic-core mbedtls)
+         #unset(MBEDTLS_INCLUDE_DIRS)
+         set(MbedTLS_FOUND ON)
+         if (PLATFORM_ESP_IDF)
+             option(BUILD_LOGREADER OFF) # Disable picolog_t build
```
TimEvens, line 104: I don't believe we need this for non-IDF builds...

src/transport_picoquic.cpp
```diff
@@ -463,7 +463,7 @@ PicoQuicTransport::Start()
     (void)picoquic_config_set_option(&config_, picoquic_option_ALPN, quicr_alpn);
     (void)picoquic_config_set_option(
       &config_, picoquic_option_CWIN_MIN, std::to_string(tconfig_.quic_cwin_minimum).c_str());
-    (void)picoquic_config_set_option(&config_, picoquic_option_MAX_CONNECTIONS, "10000");
+    (void)picoquic_config_set_option(&config_, picoquic_option_MAX_CONNECTIONS, "1");
```
TimEvens, line 466: Why 1? Has something changed where that won't limit the relay? This probably should be set in transport config.

GhostofCookie: Ah, yet another thing that needs config. Poor esp didn't have enough memory for picoquic allocating so many connections.

dependencies/CMakeLists.txt
GhostofCookie: I think it's fairly idiomatic either way?

cmake/FindMbedTLS.cmake
```diff
@@ -2,6 +2,10 @@
 # SPDX-License-Identifier: BSD-2-Clause
 
 find_path(MBEDTLS_INCLUDE_DIRS mbedtls/ssl.h)
```
BrettRegnier, line 4: Since it was find_path(MBEDTLS_INCLUDE_DIRS mbedtls/ssl.h mbedtls/esp_config.h)

Should we do
SET(MBED_HEADERS mbedtls/esp_config.h)
if (PLATFORM_ESP_IDF)
list(APPEND MBED_HEADERS mbedtls/esp_config.h)
endif() 
find_path(MBEDTLS_INCLUDE_DIRS MBED_HEADERS)

PR #539, opened 2025-04-19
Additional fixes for multi-publisher handling

TimEvens:


src/transport.cpp
```diff
@@ -628,6 +628,7 @@ namespace quicr {
     void Transport::SubscribeTrack(TransportConnId conn_id, std::shared_ptr<SubscribeTrackHandler> track_handler)
     {
         const auto& tfn = track_handler->GetFullTrackName();
+        track_handler->connection_handle_ = conn_id;
```
TimEvens, line 631: This was missing, which caused the track handler to have a connection handle/ID of zero, which broke cleanup on subscribe error.  Super simple fix. :)

src/transport.cpp
```diff
@@ -853,6 +854,9 @@ namespace quicr {
         } else {
             auto pub_n_it = pub_ns_it->second.find(th.track_name_hash);
             if (pub_n_it == pub_ns_it->second.end()) {
+                track_handler->SetStatus(pub_ns_it->second.begin()->second->GetStatus());
+                SendAnnounce(conn_it->second, tfn.name_space);
```
TimEvens, line 858: We now send duplicate announces, one for each publish track.  The reason is that we need the relay to be notified that there is a new publish track and a subscribe that may have errored last time should be resent now.

GhostofCookie (approved):

PR #553, opened 2025-04-29
Draft 11 Control Messages

RichLogan:


suhasHere (approved):
merging this PR in .. It will not compile as-is, but we are ok to fail for a bit before this stabilizes.

TimEvens:
How about we do our PRs and get them ready, then let's merge/squash this first followed by each subsequent PR based on their readiness.

TimEvens:
Created issue #560 because this PR removed our custom NewGroup message.  This is going to be a challenge to use the parser to regenerate control message parsing/handling when we also have custom messages.

RichLogan:
@TimEvens Scott already has a mechanism for this, I regenerate with that now.

PR #554, opened 2025-04-30
Add draft 11 subgroup header + logic

### LINKED ISSUE #550: Implement different types of SubGroupHeader
Draft-11 introduced different types of SubGroup header for optimization purposes. See https://moq-wg.github.io/moq-transport/draft-ietf-moq-transport.html#name-subgroup-header

RichLogan:
Closes #550 

We could better represent the modes in a fancier way if we want. I would also prefer that setting the serialisation flag of extensions if enforced at construction time, but this would require some changes in the stream buffer to forward emplace arguments if so, so I opted to defer that.

### DIFF
--- a/include/quicr/detail/messages.h
+++ b/include/quicr/detail/messages.h
@@ -19,13 +19,108 @@ namespace quicr::messages {
     using ObjectPriority = uint8_t;
     using Extensions = std::map<uint64_t, Bytes>;
 
+    /// @brief Stream Header types.
+    enum class StreamHeaderType : uint8_t
+    {
+        kZeroNoExtensions = 0x08,          // No extensions, Subgroup ID = 0
+        kZeroWithExtensions = 0x09,        // With extensions, Subgroup ID = 0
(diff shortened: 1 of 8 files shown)


include/quicr/detail/messages.h
```diff
@@ -19,14 +19,109 @@ namespace quicr::messages {
...
+        kExplicitWithExtensions = 0x0D,    // With extensions, Explicit Subgroup ID
+    };
+
+    /**
+     * Check if this Stream Header type will serialize extensions (even empty ones).
+     * @param type Type to check.
+     * @return True if the type will serialize extensions (even empty ones), false otherwise.
+     */
+    [[maybe_unused]]
+    static bool typeWillSerializeExtensions(const StreamHeaderType type)
```
GhostofCookie, line 39: Should all these methods be PascalCase?

GhostofCookie (approved):

include/quicr/detail/messages.h
RichLogan: Too much Swift. Wonder why `clang-format` doesn't care, maybe because it's static?

GhostofCookie (approved):

PR #558, opened 2025-04-30
Change control message to use 16bit payload length instead of var-int

### LINKED ISSUE #552: Control message lengths are now 16 bits
See 
```
MOQT Control Message {
  Message Type (i),
  Message Length (16),
  Message Payload (..),
}
```

TimEvens:
Resolves #552


TimEvens:
wrong branch

PR #559, opened 2025-04-30
Change control message payload size to 16bit

### LINKED ISSUE #552: Control message lengths are now 16 bits

TimEvens:
Resolves #552


src/transport.cpp
```diff
@@ -1250,33 +1250,32 @@ namespace quicr {
                     }
 
                     // Decode control payload length in bytes
-                    auto uv_sz = UintVar::Size(conn_ctx.ctrl_msg_buffer.front());
-
-                    if (conn_ctx.ctrl_msg_buffer.size() < uv_sz) {
+                    if (conn_ctx.ctrl_msg_buffer.size() < 2) {
```
RichLogan, line 1253: Perhaps this `2` could be `sizeof(payload_len)` as well so it's all consistent and dependent on the type?

GhostofCookie: +1

GhostofCookie (approved):

src/transport.cpp
TimEvens: updated.

PR #568, opened 2025-05-13
Added SSL keylog, for easier tracing with wireshark

Schweitzee:
This pull request introduces changes to enhance debugging capabilities for the `PicoQuicTransport` by enabling SSL key logging. The updates include configuring SSL key logging in the PicoQUIC library and setting the key log file from the environment.

Format is the same  as in picoquic: SSLKEYLOGFILE=<path_to_keylogfile>


RichLogan:
@Schweitzee thanks for your contribution!

I think perhaps we should add a corresponding libquicr transport config option that needs to be set for this to turned on, so nobody is in a position to leak keys when they don't mean to. I know there's the env variable, but I see the picoquic docs themselves suggest not relying on it.

RichLogan (approved):
Sorry for the delay in getting back to you on this, thanks again for your contribution!

PR #571, opened 2025-06-09
Update tconfig parameter

RichLogan:
The previous value of `1000` doesn't work as a default because of control messages using a hardcoded TTL value of `2000`.


TimEvens (approved):
1000 was working. Would be good to come back to this, but for now LGTM.

PR #572, opened 2025-06-09
Integration tests

RichLogan:
Integration test framework + initial connection test.

Could use promise/future for some of these callback checks, but right now I'm not sure they would add that much. Maybe when the tests get more complicated.


test/integration_test/test_client.h
```diff
@@ -0,0 +1,17 @@
...
+
+namespace quicr_test {
+    class TestClient final : public quicr::Client
+    {
+      public:
+        explicit TestClient(const quicr::ClientConfig& cfg);
+
+        // Connection.
+        using ClientConnected = std::function<void(const quicr::ServerSetupAttributes&)>;
+        void SetClientConnected(const ClientConnected& cb);
```
GhostofCookie, line 11: Might mention in the name that it's an event or callback.

GhostofCookie (approved):
Looks good, though it'd be good form to add newlines to the bottoms of these files.

RichLogan:
@GhostofCookie Just for you: https://github.com/Quicr/libquicr/pull/574

PR #573, opened 2025-06-11
Autogenerate ctrl_messages as part of the build

RichLogan:
Makes the draft parser generation part of the libquicr build. Maybe this is overcomplicating things, but the idea here is:

- Remove autogenerated code from the source tree
- Make it more explicit about what is generated and what is not
- Ensure that a build is always in sync with the given draft used


shenning00 (approved):
Looks fine with me.

PR #578, opened 2025-06-18
Produce docs for all detail

RichLogan:
It's possible we intentionally didn't have all of `detail` generating docs, but at least some things that we probably do want to document are missing (I was looking for `TransportConfig`), so I err'd on the side of just listing the whole directory.

I also noticed that the single same line comment was causing Doxygen to produce comments that were off-by-one for the field they matched for some of these structures (although not all of them). Moving the doc comment to above the field fixed that.

### DIFF
--- a/Doxyfile
+++ b/Doxyfile
@@ -953,8 +953,7 @@ WARN_LOGFILE           =
 # Note: If this tag is empty the current directory is searched.
 
 INPUT                  =  docs/api-main.md \
-                          include/quicr/detail/transport.h \
-                          include/quicr/detail/uintvar.h \
+                          include/quicr/detail \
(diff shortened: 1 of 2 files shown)


TimEvens:
We also talked about seeing if we can get AI to generate the API docs instead of doxygen.

RichLogan:
@TimEvens That makes sense - I kind of expected that. To be honest the real driver of this was me debugging the max connections being 1 and getting `SERVER BUSY` when connecting 2 clients. That was a non-obvious error to track down - perhaps I can add some better logging around that.

TimEvens:
@RichLogan , let's talk about this next week.

PR #581, opened 2025-06-30
Fixup byte swapping

RichLogan:
Somewhere along the way our endianness swapping got confused for the fixed 16 bit network order control message lengths.

### DIFF
--- a/src/ctrl_message_types.cpp
+++ b/src/ctrl_message_types.cpp
@@ -31,9 +31,8 @@ namespace quicr::messages {
 
     Bytes& operator<<(Bytes& buffer, std::uint16_t value)
     {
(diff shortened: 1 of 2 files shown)


GhostofCookie (approved):
Great catch!

TimEvens:
@RichLogan , @GhostofCookie , we'll need to catch this better in the future. Looks like it might have been a bad merge with a Suhas commit https://github.com/Quicr/libquicr/pull/553/commits/1f05dc2f66de917c09c9b9c89d79b984d4438a72#diff-5f18a1292c45a05d520ffb6c12e15ba0c26e2a0f8ba72f5590f870466d128e3e

PR #584, opened 2025-07-07
Add fuzzer and uintvar test

RichLogan:
Add a libfuzzer framework, with uintvar initial parser test.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -42,6 +42,7 @@ option(QUICR_BUILD_SHARED "Build quicr as a SHARED library" OFF)
 option(PLATFORM_ESP_IDF "Enabble support for esp-idf (Default OFF)" OFF)
 option(USE_MBEDTLS OFF)
 option(DRAFT_PARSER_SETUP_VENV "Set up Python virtual environment for draft parser" ON)
+option(QUICR_BUILD_FUZZ "Build fuzzer targets" OFF)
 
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -107,6 +107,13 @@ if (BUILD_BENCHMARKING AND quicr_BUILD_BENCHMARKS)
    add_subdirectory(benchmark)
 endif()
 
+###
+### Fuzzing targets
+###
+if(QUICR_BUILD_FUZZ)
+   add_subdirectory(fuzz)
+endif()
```
GhostofCookie, line 115: Might be good to add
```cmake
option(QUICR_BUILD_FUZZ OFF)
```
near the top? Just to help with enumerating what options are available

fuzz/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 Cisco Systems
+# SPDX-License-Identifier: BSD-2-Clause
+
+if(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+    option(QUICR_BUILD_FUZZ "Build fuzzer targets" OFF)
+    if(QUICR_BUILD_FUZZ)
```
GhostofCookie, line 6: I think these should be handled only by the root cmake. The only way this cmake file gets called is if the root file has passed this check.

PR #591, opened 2025-07-14
Test of group_id time queue

GhostofCookie:


src/transport.cpp
```diff
@@ -1206,7 +1207,7 @@ namespace quicr {
         }
 
         auto result = quic_transport_->Enqueue(
-          track_handler.connection_handle_, track_handler.publish_data_ctx_id_, data, priority, ttl, 0, eflags);
+          track_handler.connection_handle_, track_handler.publish_data_ctx_id_, 0, data, priority, ttl, 0, eflags);
```
TimEvens, line 1210: `SendData()` is used with pipelining.  We need to talk about this a little more because pipelining is group unaware... Setting it to group 0 will result in not purging all related objects on expiry. Instead, it'll empty the whole queue of pipeline data.

TimEvens (commented):
I think this breaks for pipelining.  We need to figure out the best way to handle that.

include/quicr/detail/quic_transport.h
```diff
@@ -419,16 +419,18 @@ namespace quicr {
          *
          * @param[in] context_id	Identifying the connection
          * @param[in] data_ctx_id	stream Id to send data on
-         * @param[in] bytes		Data to send/write
-         * @param[in] priority          Priority of the object, range should be 0 - 255
-         * @param[in] ttl_ms            The age the object should exist in queue in milliseconds
-         * @param[in] delay_ms          Delay the pop by millisecond value
-         * @param[in] flags             Flags for stream and queue handling on enqueue of object
+         * @param[in] groupd_id     The id of the group we're currently enqueuing.
```
RichLogan, line 422: `groupd_id` -> `group_id`

src/transport.cpp
TimEvens: PR #605 resolves this.

PR #592, opened 2025-07-15
Initial support for delivery timeout.

GhostofCookie:


include/quicr/detail/ctrl_message_types.h
```diff
@@ -118,15 +118,27 @@ namespace quicr::messages {
...
+        kMaxRequestId = 0x02, // version specific, unused
+        kEndpointId = 0xF1,   // Endpoint ID, using temp value for now
+        kInvalid = 0xFF,      // used internally.
     };
 
-    using Parameter = KeyValuePair<ParameterType>;
+    enum struct VersionSpecificParameterType : uint64_t
+    {
+        kDeliveryTimeout = 0x02,
+        kAuthorizationToken = 0x03,
```
RichLogan, line 132: I see that this one is kind of weird to talk about, in MoQ they say this is setup, but actually can appear basically everywhere (CLIENT_SETUP, SERVER_SETUP, SUBSCRIBE, SUBSCRIBE_NAMESPACE, ANNOUNCE, TRACK_STATUS or FETCH). Maybe we should track such a thing one in yet another enum. Or in both. moq pls.

include/quicr/detail/ctrl_message_types.h
```diff
@@ -118,15 +118,27 @@ namespace quicr::messages {
...
+        kMaxCacheDuration = 0x04,
+        kInvalid = 0xFF, // used internally.
+    };
+
+    using ParameterType [[deprecated]] = SetupParameterType;
```
RichLogan, line 137: This is more work for you, but I'd say just nuke these things rather than leave them around as deprecated. Not sure we care about backwards compatibility. Can open an issue and do it later though.

GhostofCookie: @shenning00

PR #594, opened 2025-07-18
Draft12 publish

### LINKED ISSUE #400: pause/resume - lib quicr

TimEvens:
Adds support for PUBLISH while supporting backwards compatibility using ANNOUNCE flow. 

This PR also adds support for pause/resume in subscribe handler. Resolves #400 

Bumps version to draft 12

### DIFF
--- a/include/quicr/subscribe_track_handler.h
+++ b/include/quicr/subscribe_track_handler.h
@@ -63,17 +63,20 @@ namespace quicr {
          *
          * @param full_track_name           Full track name struct
          * @param joining_fetch             If set, subscribe with a joining fetch using these attributes.
+         * @param publisher_initiated       True if publisher initiated the subscribe, otherwise False
          */
         SubscribeTrackHandler(const FullTrackName& full_track_name,
                               messages::SubscriberPriority priority,
                               messages::GroupOrder group_order,
                               messages::FilterType filter_type,
-                              const std::optional<JoiningFetch>& joining_fetch = std::nullopt)
+                              const std::optional<JoiningFetch>& joining_fetch = std::nullopt,
+                              bool publisher_initiated = false)
           : BaseTrackHandler(full_track_name)
           , priority_(priority)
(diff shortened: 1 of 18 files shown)


include/quicr/subscribe_track_handler.h
```diff
@@ -63,17 +63,20 @@ namespace quicr {
...
                               messages::FilterType filter_type,
-                              const std::optional<JoiningFetch>& joining_fetch = std::nullopt)
+                              const std::optional<JoiningFetch>& joining_fetch = std::nullopt,
+                              bool publisher_initiated = false)
           : BaseTrackHandler(full_track_name)
           , priority_(priority)
           , group_order_(group_order)
           , filter_type_(filter_type)
           , joining_fetch_(joining_fetch)
+          , publisher_initiated_(publisher_initiated)
```
RichLogan, line 79: Just an API design thing, but should we enforce the fact that a publisher initiated subscribe handler can't use joining fetch? Or maybe two constructors?

src/client.cpp
```diff
@@ -543,6 +543,50 @@ namespace quicr {
...
                 return true;
             }
+
+            case messages::ControlMessageType::kPublishOk: {
+                messages::PublishOk msg(
+                  [](messages::PublishOk& msg) {
+                      if (msg.filter_type == messages::FilterType::kAbsoluteStart ||
+                          msg.filter_type == messages::FilterType::kAbsoluteRange) {
+                          msg.group_0 =
+                            std::make_optional<messages::PublishOk::Group_0>(messages::PublishOk::Group_0{ 0, 0 });
```
RichLogan, line 553: Assuming we're going to come back to these (should we make a note?)

include/quicr/subscribe_track_handler.h
TimEvens: Changed it to null opt joining fetch if provided when publisher initiated is true.

PR #595, opened 2025-07-18
Support KVP + length based extensions

RichLogan:
Extensions are now a series of MoQ key-value-pairs, with the series prefixed with the byte count of the whole extensions block, rather than the count of extensions serialised.

### DIFF
--- a/src/messages.cpp
+++ b/src/messages.cpp
@@ -21,57 +21,13 @@ namespace quicr::messages {
     }
 
     template<class StreamBufferType>
-    static bool ParseExtensions(StreamBufferType& buffer,
-                                uint64_t& count,
-                                std::optional<Extensions>& extensions,
-                                std::optional<uint64_t>& current_tag)
+    static bool ReadUintVField(StreamBufferType& buffer, UintVar& field)
     {
-
-        // get the count of extensions
-        if (count == 0 && !ParseUintVField(buffer, count)) {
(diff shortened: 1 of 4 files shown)


src/messages.cpp
```diff
@@ -329,8 +373,18 @@ namespace quicr::messages {
                 [[fallthrough]];
             }
             case 5: {
-                if (!ParseExtensions(buffer, msg.num_extensions, msg.extensions, msg.current_tag)) {
-                    return false;
+                assert(msg.type.has_value());
```
TimEvens, line 376: Let's throw instead of assert....

src/messages.cpp
```diff
@@ -538,8 +596,12 @@ namespace quicr::messages {
         buffer << UintVar(msg.object_id);
         assert(msg.stream_type.has_value()); // Stream type must have been set before serialization.
         const auto properties = StreamHeaderProperties(*msg.stream_type);
+        if (!properties.may_contain_extensions && msg.extensions.has_value()) {
+            // This is not allowed.
+            assert(false);
```
TimEvens, line 601: throw instead of assert.

TimEvens (approved):
Seems like both `extensions_bytes_left` and `extensions_bytes_remaining` are used for the same thing. Would make sense to change **_left** to **_remaining**

Looks good. Left a few comments.

PR #611, opened 2025-07-22
Add support for publisher to use their own track alias, other fixes

TimEvens:
This includes the below other fixes and updates:
* Fixes request_id increment based on messages received
* Changes publisher GetTrackAlias and adds SetTrackAlias to be consistent with subscribe track handler
* Default will generate track alias, but user can set a different one
* fixes typos
* Changes state tracking name to sub_by_recv_track_alias.
* Update state tracking sub_by_recv_track_alias entries.  This is the publisher received alias to the connection. This allows subscribe handler lookup using publisher assigned track alias
* Add received track alias to subscribe handler


cmd/examples/server.cpp
```diff
@@ -892,7 +889,7 @@ class MyServer : public quicr::Server
                                 conn_h,
                                 th.track_fullname_hash,
                                 track_alias);
-                    qserver_vars::pub_subscribes[track_alias][conn_h] = copy_sub_track_h;
+                    qserver_vars::pub_subscribes[1][conn_h] = copy_sub_track_h;
```
GhostofCookie: Magic 1

GhostofCookie (approved):

PR #615, opened 2025-07-24
Fix subscribe ok

RichLogan:
Fix bug in `SUBSCRIBE_OK` message construction, where content exists having a value of `0` would result in a location of `{0, 0}` being encoded, rather than no location at all. I have changed the API so that the type system stops a bad combo of arguments. 

However, I think this is a good example of how something like #607 would allow us to encode these rules that aren't otherwise parsed from the spec and catch them as early as possible in deserialize, and also as a last chance check on serialize. 

Some known message vectors would also be great test suite addition - and probably a good addition as a draft appendix?


TimEvens (approved):
LGTM

PR #616, opened 2025-07-24
Fixup some other optionals

RichLogan:
Found some other cases where badly constructed messages could theoretically be sent - this will throw in those cases (but we still need to handle the actual logic).


TimEvens (approved):
LGTM

Better to throw an error for now than to silently accept a filter type that we don't implement.

PR #618, opened 2025-07-24
Correct FilterType naming / values

RichLogan:
Noticed these names and values weren't matching v12. (The `{2,1,3,4}` order was intentional to match the order they're described in the draft). 

>   There are 4 types of filters:
> 
>    Largest Object (0x2): The filter Start Location is {Largest
>    Object.Group, Largest Object.Object + 1} and Largest Object is
>    communicated in SUBSCRIBE_OK.  If no content has been delivered yet,
>    the filter Start Location is {0, 0}. There is no End Group - the
>    subscription is open ended.  Note that due to network reordering or
>    prioritization, relays can receive Objects with Locations smaller
>    than Largest Object after the SUBSCRIBE is processed, but these
>    Objects do not pass the Largest Object filter.
> 
>    Next Group Start (0x1): The filter start Location is {Largest
>    Object.Group + 1, 0} and Largest Object is communicated in
>    SUBSCRIBE_OK.  If no content has been delivered yet, the filter Start
>    Location is {0, 0}.  There is no End Group - the subscription is open
>    ended.  For scenarios where the subscriber intends to start more than
>    one group in the future, it can use an AbsoluteStart filter instead.
> 
>    AbsoluteStart (0x3): The filter Start Location is specified
>    explicitly in the SUBSCRIBE message.  The Start specified in the
>    SUBSCRIBE message MAY be less than the Largest Object observed at the
>    publisher.  There is no End Group - the subscription is open ended.
>    To receive all Objects that are published or are received after this
>    subscription is processed, a subscriber can use an AbsoluteStart
>    filter with Start = {0, 0}.
> 
>    AbsoluteRange (0x4): The filer Start Location and End Group are
>    specified explicitly in the SUBSCRIBE message.  The Start specified
>    in the SUBSCRIBE message MAY be less than the Largest Object observed
>    at the publisher.  If the specified End Group is the same group
>    specified in Start, the remainder of that Group passes the filter.
>    End Group MUST specify the same or a larger Group than specified in
>    Start.
>

### DIFF
--- a/include/quicr/detail/ctrl_message_types.h
+++ b/include/quicr/detail/ctrl_message_types.h
@@ -212,8 +212,8 @@ namespace quicr::messages {
 
     enum struct FilterType : uint64_t
     {
-        kNextGroupStart = 0x2,
-        kLargestObject = 0x1,
+        kLargestObject = 0x2,
+        kNextGroupStart = 0x1,
         kAbsoluteStart = 0x3,
         kAbsoluteRange = 0x4
     };
--- a/include/quicr/detail/ctrl_message_types.h
+++ b/include/quicr/detail/ctrl_message_types.h
@@ -212,11 +212,10 @@ namespace quicr::messages {
 
     enum struct FilterType : uint64_t
     {
-        kNone = 0x0,
-        kLatestGroup,
-        kLatestObject,
-        kAbsoluteStart,
-        kAbsoluteRange
+        kNextGroupStart = 0x2,
+        kLargestObject = 0x1,
+        kAbsoluteStart = 0x3,
+        kAbsoluteRange = 0x4
     };
 
     enum class TrackStatusCode : uint64_t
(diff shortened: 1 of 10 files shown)


include/quicr/detail/ctrl_message_types.h
```diff
@@ -212,11 +212,10 @@ namespace quicr::messages {
 
     enum struct FilterType : uint64_t
     {
-        kNone = 0x0,
-        kLatestGroup,
-        kLatestObject,
-        kAbsoluteStart,
-        kAbsoluteRange
+        kNextGroupStart = 0x2,
+        kLargestObject = 0x1,
```
TimEvens, line 216:
```suggestion
        kNextGroupStart = 0x1,
        kLargestObject = 0x2, 
```

TimEvens (changes requested):
Thanks for the cleanup. One enum change required based on the draft.

PR #619, opened 2025-07-25
Make draft parsing of optionals stricter

RichLogan:
- Callbacks not provided for outbound full message constructors
- Callbacks MUST be provided for inbound stream constructors


TimEvens (approved):
LGTM

PR #620, opened 2025-07-25
Remove redundant qualifiers

RichLogan:
Just removing the redundant `quicr::` qualifier from these operators.


TimEvens (approved):
Looks good, but in some cases we were being explicit with qualifiers to make it very clear when reading code.  This change does remove that, especially if `Bytes` is defined again somewhere else.

RichLogan:
@TimEvens totally fine to close this if so - I just got baited by CLion constantly greying them out :D

PR #621, opened 2025-07-25
Remove unneeded stream operators

RichLogan:
- Single templated operator for all varint backed types

This was "accidentally" already the case, because the KVP templating would pick up all stream operators for `uint64_t` backed types already. So if we're continuing with that approach, we can remove a number of existing operators.

Renamed the `concept` to make it more obvious. 

The FETCH enums were specified as `uint8_t` which means they were encoding to `(8)` not `(i)` as they should have been - but for these low values they're (luckily) equivalent. Fixed that.


TimEvens:
Revisit later. Closing due to it being out of date.

PR #631, opened 2025-08-04
Support multi-publisher and new group request

TimEvens:


src/transport.cpp
```diff
@@ -1163,9 +1161,10 @@ namespace quicr {
 
         track_handler->SetStatus(FetchTrackHandler::Status::kPendingResponse);
 
+        conn_it->second.sub_tracks_by_request_id[*track_handler->GetRequestId()] = std::move(track_handler);
```
RichLogan, line 1164: Double move / use-after-move after this line

TimEvens: Thanks. That was a bad merge of main.

PR #633, opened 2025-08-07
Fix throw catching

TimEvens:
Fixes issues causing relay to crash due to unhandled re-throws.


src/transport.cpp
```diff
@@ -1353,8 +1352,11 @@ namespace quicr {
     void Transport::RemoveAllTracksForConnectionClose(ConnectionContext& conn_ctx)
     {
         // clean up subscriber handlers on disconnect
-        for (const auto& [sub_id, handler] : conn_ctx.sub_tracks_by_request_id) {
+        for (const auto& [req_id, handler] : conn_ctx.sub_tracks_by_request_id) {
             RemoveSubscribeTrack(conn_ctx, *handler, false);
+            if (handler->GetConnectionId() == conn_ctx.connection_handle) {
```
TimEvens, line 1357: This was the root cause of the problem causing the churn of sending subscribe update (aka pause) on the same connection that was closed. It was triggered by not receiving an unsubscribe and when the publisher and subscriber for the same full track name are the same connection.

src/transport.cpp
```diff
@@ -212,7 +212,7 @@ namespace quicr {
         SendCtrlMsg(conn_ctx, buffer);
     } catch (const std::exception& e) {
         SPDLOG_LOGGER_ERROR(logger_, "Caught exception sending ClientSetup (error={})", e.what());
-        throw;
+        // TODO: add error handling in libquicr in calling function
```
GhostofCookie, line 215: Nice! I'll make a follow up PR to properly address the usage of these errors.

GhostofCookie (approved):

PR #638, opened 2025-08-14
Protect update func calls

RichLogan:
Protect these `std::function` calls with null checks. Add a test to ensure there is no crash when using a handler without a transport. I know there's discussion about rearchitecting this and this fix here is very minimal but it's still better than nothing so figured I'd throw it in.


TimEvens (approved):
Thanks for the quick fix on nullptr.  We'll need to follow up with removing when transport is destructed.

PR #642, opened 2025-08-20
Ensure shared transport

GhostofCookie:
Remove function callback in publishers for sending data, and enforce that transport MUST be shared.

### DIFF
--- a/src/publish_fetch_handler.cpp
+++ b/src/publish_fetch_handler.cpp
@@ -1,26 +1,113 @@
 // SPDX-FileCopyrightText: Copyright (c) 2025 Cisco Systems
 // SPDX-License-Identifier: BSD-2-Clause
 
-#include <quicr/publish_fetch_handler.h>
+#include "quicr/publish_fetch_handler.h"
+
+#include "quicr/detail/transport.h"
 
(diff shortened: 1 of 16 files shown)


src/publish_fetch_handler.cpp
```diff
@@ -1,26 +1,113 @@
...
+        std::uint64_t group_id = object_headers.group_id;
+        std::uint64_t object_id = object_headers.object_id;
+        std::uint16_t ttl = object_headers.ttl.has_value() ? object_headers.ttl.value() : default_ttl_;
+        std::uint8_t priority =
+          object_headers.priority.has_value() ? object_headers.priority.value() : default_priority_;
+
+        object_msg_buffer_.clear();
+
+        switch (default_track_mode_) {
+            case TrackMode::kDatagram: {
```
TimEvens, line 36: This appears to be a cut/paste of SendObject from transport. Makes sense when it was common code. Now that this is moved to each handler, this doesn't make sense here. 

Fetch does not use datagram.  Fetch does not use more than one stream.

TimEvens (approved):
LGTM.... working with LAPS as well.

PR #671, opened 2025-09-15
Add minimal support for absolute joining fetch

RichLogan:


src/server.cpp
```diff
@@ -750,10 +752,29 @@ namespace quicr {
                         }
                         largest_location = *opt_largest_location;
 
-                        // TODO(RichLogan): Check this when FETCH v11 checked.
-                        const auto start_group = msg.group_1->joining.joining_start <= largest_location.group
-                                                   ? largest_location.group - msg.group_1->joining.joining_start
-                                                   : largest_location.group;
+                        messages::GroupId start_group;
+                        switch (msg.fetch_type) {
```
TimEvens, line 756: Not ideal that fetch type is being checked a second time.

PR #672, opened 2025-09-15
Fixup FETCH data plane logic

RichLogan:
With some recent changes it looks like the FETCH data plan was using the stream subgroup header format. This fixes that up. Our fetch integration test doesn't actually roundtrip data, so didn't catch this.


TimEvens (approved):
LGTM

PR #679, opened 2025-09-16
Update subscribe update for draft-14

### LINKED ISSUE #660: Add state handling for using susbcription request_id in the subscribe_update
SubscribeUpdate has a new field called "Subscribe Request Id" that refers to the subscription being updated. Update the implementation to use the new parameter and for updates to use its own request id.

TimEvens:
* Change subscribe update to use its own request id
* Use the update subscribe request id to find the subscribe that the update intends to update

Resolves #660


include/quicr/detail/transport.h
```diff
@@ -327,6 +328,11 @@ namespace quicr {
         static constexpr std::size_t kControlMessageBufferSize = 4096;
         struct ConnectionContext
         {
+            ConnectionContext(const ConnectionContext& other)
+              : next_request_id(other.next_request_id.load(std::memory_order_seq_cst))
```
GhostofCookie, line 332: If the order is seq, then may has well just set the copy constructor to default, since that's the default for atomic

GhostofCookie (approved):

PR #680, opened 2025-09-17
Add basic client FETCH protocol support

RichLogan:
Stub in some basic protocol support for relative FETCH to the client. This is effectively the same as what we have in server, so I do think this can be refactored up to transport and shared between the two, but this gets us to protocol interop I think.

### DIFF
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -714,8 +714,8 @@ namespace quicr {
                       }
                   },
                   [](messages::Fetch& msg) {
-                      // TODO: Add support for absolute joining fetch
-                      if (msg.fetch_type == messages::FetchType::kRelativeJoiningFetch) {
+                      if (msg.fetch_type == messages::FetchType::kRelativeJoiningFetch ||
+                          msg.fetch_type == messages::FetchType::kAbsoluteJoiningFetch) {
                           msg.group_1 = std::make_optional<messages::Fetch::Group_1>();
                       }
                   });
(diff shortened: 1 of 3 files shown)


src/client.cpp
```diff
@@ -733,7 +733,50 @@ namespace quicr {
...
+                                           msg.request_id,
+                                           messages::FetchErrorCode::kInvalidJoiningRequestId,
+                                           "Joining track not found");
+                            return true;
+                        }
+                        const auto location = sub->second->GetLatestLocation();
+                        if (!location.has_value()) {
+                            SendFetchError(conn_ctx,
+                                           msg.request_id,
+                                           messages::FetchErrorCode::kInvalidRange,
```
TimEvens, line 759: Should this be `kNoObjects`

src/client.cpp
```diff
@@ -733,7 +733,50 @@ namespace quicr {
...
+                                           "No objects published");
+                            return true;
+                        }
+                        messages::GroupId start_group;
+                        if (msg.fetch_type == messages::FetchType::kAbsoluteJoiningFetch) {
+                            start_group = joining.joining_start;
```
TimEvens, line 765: Draft 14  says: 

> For an Absolute Joining Fetch, the publisher sets the Start Location to Joining Start.

This would include object as well.  For relative, it does knind of indicate that object is zero based on `{Subscribe Largest Location.Group - Joining Start, 0}`

TimEvens (commented):
Looks good but a few comments.

PR #685, opened 2025-09-19
Fix gap extension double encode

RichLogan:
- Gap extensions were being double VarInt encoded. Extensions expects non-varint `std::uint64_t`s that it will compress on serialisation when the key is even. I would very much like to refactor to extensions to more strongly type this even/odd capability because this is an easy mistake to make. 
- Added a test that crashed on the bad behaviour, as well as real data roundtrip test, but this requires a test framework that can actually roundtrip data which we don't have yet, but I'd already written it so figured I'd leave it here.

### DIFF
--- a/src/publish_track_handler.cpp
+++ b/src/publish_track_handler.cpp
@@ -169,23 +169,27 @@ namespace quicr {
         auto object_extensions = object_headers.extensions;
 
         if (group_id_delta > 1) {
-            auto value = UintVar(group_id_delta - 1);
+            const std::uint64_t value = group_id_delta - 1;
+            std::vector<std::uint8_t> value_bytes(sizeof(value));
+            memcpy(value_bytes.data(), &value, sizeof(value));
             if (not object_extensions.has_value()) {
                 object_extensions = Extensions{};
             }
 
             object_extensions->try_emplace(static_cast<uint64_t>(messages::ExtensionHeaderType::kPriorGroupIdGap),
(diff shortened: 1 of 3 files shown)


src/publish_track_handler.cpp
```diff
@@ -169,23 +169,27 @@ namespace quicr {
         auto object_extensions = object_headers.extensions;
 
         if (group_id_delta > 1) {
-            auto value = UintVar(group_id_delta - 1);
+            const std::uint64_t value = group_id_delta - 1;
+            std::vector<std::uint8_t> value_bytes(sizeof(value));
+            memcpy(value_bytes.data(), &value, sizeof(value));
```
TimEvens, line 174: we should not be doing memcpy here.

TimEvens (commented):
varint returns the encoded bytes from uint64. If that isn't working, then that needs to be fixed. We should not be doing memcpy here to get the vector of bytes from varint class.

TimEvens (changes requested):
See https://github.com/Quicr/libquicr/blob/main/test/uintvar.cpp#L63, which is supposed to work.  That was being used and if it's broken, let's fix that.

TimEvens (approved):
I'll merge this as a temp break fix, but the issue is with serialization and the extensions map.  While it seems to be nice that serialization does the varint encoding from a vector, it still requires assignment to encode host order as a vector, which is not necessary.

PR #696, opened 2025-09-24
Remove fake max for priorities and limit to size of uint8_t.

### LINKED ISSUE #695: Change priority queue max to support values 0 - 255
Priority queue was updated to use a vector instead of array, so the PMAX is no longer needed to restrict the value. 

Review and fix picoquic priority mappings.

GhostofCookie:
Closes #695


TimEvens (approved):
LGTM

PR #698, opened 2025-09-25
Switch to use picoquic network thread api

### LINKED ISSUE #697: Use picoquic network thread API to improve latency
Currently marking active is delayed till loop timeout, which is by default 500us.  This seems small, but with a chain of relays, this adds up.  Switching to the picoquic network thread API should resolve this.

TimEvens:
This moves from libquicr created picoquic thread to picoquic creating the thread instead.  The advantage with this is that picoquic enables faster loop notifications than having to run the loop in a fast 500us loop.  This performs better and improves latency.

The changes also support adding multiple clients using the same quic instance. This will help dramatically with peering so that each outbound peer connection is not creating a new quic context. Another PR will expand on the client changes using network thread to expose a connect and disconnect API with the client.   

Resolves #697

### DIFF
--- a/src/transport_picoquic.cpp
+++ b/src/transport_picoquic.cpp
@@ -1665,6 +1665,7 @@ PicoQuicTransport::Server()
       picoquic_start_network_thread(quic_ctx_, &quic_network_thread_params_, PqLoopCb, this, &quic_loop_return_value_);
 
     if (quic_ctx_ == NULL || quic_network_thread_ctx_ == NULL) {
+        SPDLOG_LOGGER_ERROR(logger, "Failed to start picoquic network thread");
         picoquic_free(quic_ctx_);
         quic_ctx_ = NULL;
         SetStatus(TransportStatus::kShutdown);
@@ -1672,7 +1673,7 @@ PicoQuicTransport::Server()
 
     // Wait for something to happen with the thread
     while (!quic_network_thread_ctx_->thread_is_ready && !quic_network_thread_ctx_->return_code) {
-        std::this_thread::sleep_for(std::chrono::milliseconds(1));
+        std::this_thread::sleep_for(std::chrono::milliseconds(100));
     }
 
     if (quic_network_thread_ctx_->return_code) {
--- a/src/transport_picoquic.cpp
+++ b/src/transport_picoquic.cpp
@@ -1657,7 +1657,7 @@ PicoQuicTransport::Server()
     quic_network_thread_params_.local_port = serverInfo_.port;
     quic_network_thread_params_.local_af = PF_UNSPEC;
     quic_network_thread_params_.dest_if = 0;
-    quic_network_thread_params_.socket_buffer_size = 2000000;
+    quic_network_thread_params_.socket_buffer_size = tconfig_.socket_buffer_size;
     quic_network_thread_params_.do_not_use_gso = 0;
 
     SPDLOG_LOGGER_DEBUG(logger, "Starting picoquic network thread");
@@ -1673,7 +1673,7 @@ PicoQuicTransport::Server()
 
     // Wait for something to happen with the thread
     while (!quic_network_thread_ctx_->thread_is_ready && !quic_network_thread_ctx_->return_code) {
-        std::this_thread::sleep_for(std::chrono::milliseconds(100));
+        std::this_thread::sleep_for(std::chrono::milliseconds(1));
     }
 
     if (quic_network_thread_ctx_->return_code) {
@@ -1682,8 +1682,6 @@ PicoQuicTransport::Server()
         SetStatus(TransportStatus::kShutdown);
         return;
     }
-
-    SetStatus(TransportStatus::kReady);
 }
 
 TransportConnId
--- a/src/transport_picoquic.cpp
+++ b/src/transport_picoquic.cpp
@@ -1701,7 +1701,7 @@ PicoQuicTransport::StartClient()
     std::unique_lock lock(mtx);
 
     picoquic_runner_queue_.Push([this, &conn_id, &cv, &mtx]() {
-        auto notifyCaller = [&cv, &conn_id, &mtx](uint64_t id) {
+        auto notify_caller = [&cv, &conn_id, &mtx](uint64_t id) {
             std::lock_guard _(mtx);
             conn_id = id;
 
@@ -1718,7 +1718,7 @@ PicoQuicTransport::StartClient()
         if (ret != 0 || server_address.ss_family == 0) {
             SPDLOG_LOGGER_ERROR(
               logger, "Failed to resolve server: {0} port: {1}", serverInfo_.host_or_ip, serverInfo_.port);
-            notifyCaller(1);
+            notify_caller(1);
             return 0;
         }
 
@@ -1738,7 +1738,7 @@ PicoQuicTransport::StartClient()
 
         if (cnx == nullptr) {
             SPDLOG_LOGGER_ERROR(logger, "Could not create picoquic connection client context");
-            notifyCaller(1);
+            notify_caller(1);
 
             return PICOQUIC_ERROR_DISCONNECTED;
         }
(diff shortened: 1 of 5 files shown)


src/transport_picoquic.cpp
```diff
@@ -1629,121 +1660,172 @@ PicoQuicTransport::CheckConnsForCongestion()
...
+    cv.wait_for(lock, std::chrono::milliseconds(3000), [&]() { return conn_id > 0; });
 
-    if (conn_ctx == nullptr) {
-        SPDLOG_LOGGER_ERROR(logger, "Client connection does not exist, check connection settings.");
+    SPDLOG_LOGGER_DEBUG(logger, "Got client connection context conn_id: {}", conn_id);
+    if (conn_id <= 1) {
+        SPDLOG_LOGGER_DEBUG(logger, "Client connection to {}:{} failed", serverInfo_.host_or_ip, serverInfo_.port);
         SetStatus(TransportStatus::kDisconnected);
-        return;
+        return 0;
```
GhostofCookie, line 1782: I don't think I've seen any of the callback functions return anything other than 0, so what is this for?

src/transport_picoquic.cpp
```diff
@@ -622,7 +629,11 @@ PicoQuicTransport::Enqueue(const TransportConnId& conn_id,
         if (!conn_ctx_it->second.mark_dgram_ready) {
             conn_ctx_it->second.mark_dgram_ready = true;
 
-            picoquic_runner_queue_.Push([this, conn_id]() { MarkDgramReady(conn_id); });
+            picoquic_runner_queue_.Push([this, conn_id]() {
+                MarkDgramReady(conn_id);
+                return 0;
+            });
+            picoquic_wake_up_network_thread(quic_network_thread_ctx_);
```
GhostofCookie, line 636: Seems like we should make wrapper for anytime we push to the runner queue we should call wake up afterwards

src/transport_picoquic.cpp
```diff
@@ -985,24 +1001,29 @@ PicoQuicTransport::CreateDataContextBiDirRecv(TransportConnId conn_id, uint64_t
...
+        return 0;
     }
 
     // note: check before running move of optional, which is more CPU taxing when empty
     while (auto cb = picoquic_runner_queue_.Pop()) {
         try {
-            (*cb)();
+            if (auto ret = (*cb)()) {
+                SPDLOG_LOGGER_ERROR(logger, "PQ function resulted in error: {}", ret);
+                return ret;
```
GhostofCookie, line 1017: All the callbacks so far return 0 explicitly, so why do we do this?

src/transport_picoquic.cpp
```diff
@@ -1210,9 +1239,13 @@ PicoQuicTransport::SendStreamBytes(DataContext* data_ctx, uint8_t* bytes_ctx, si
...
 
-            picoquic_runner_queue_.Push([this, conn_id = data_ctx->conn_id, data_ctx_id = data_ctx->data_ctx_id]() {
-                MarkStreamActive(conn_id, data_ctx_id);
-            });
+            if (!data_ctx->tx_data->Empty()) {
+                picoquic_runner_queue_.Push([this, conn_id = data_ctx->conn_id, data_ctx_id = data_ctx->data_ctx_id]() {
+                    MarkStreamActive(conn_id, data_ctx_id);
+                    picoquic_wake_up_network_thread(quic_network_thread_ctx_);
+                    return 0;
+                });
```
GhostofCookie, line 1247: Are we missing a call to wake up the thread?

src/transport_picoquic.cpp
```diff
@@ -1629,121 +1660,172 @@ PicoQuicTransport::CheckConnsForCongestion()
 void
 PicoQuicTransport::Server()
 {
-    int ret = picoquic_packet_loop(quic_ctx_, serverInfo_.port, PF_UNSPEC, 0, 2000000, 0, PqLoopCb, this);
 
-    if (quic_ctx_ != NULL) {
+    quic_network_thread_params_.local_port = serverInfo_.port;
+    quic_network_thread_params_.local_af = PF_UNSPEC;
+    quic_network_thread_params_.dest_if = 0;
+    quic_network_thread_params_.socket_buffer_size = 2000000;
```
GhostofCookie, line 1667: Looks magic, potential for a config variable?

src/transport_picoquic.cpp
```diff
@@ -1629,121 +1660,172 @@ PicoQuicTransport::CheckConnsForCongestion()
...
+    if (quic_ctx_ == NULL || quic_network_thread_ctx_ == NULL) {
         picoquic_free(quic_ctx_);
         quic_ctx_ = NULL;
+        SetStatus(TransportStatus::kShutdown);
     }
 
-    SPDLOG_LOGGER_INFO(logger, "picoquic packet loop ended with {0}", ret);
+    // Wait for something to happen with the thread
+    while (!quic_network_thread_ctx_->thread_is_ready && !quic_network_thread_ctx_->return_code) {
+        std::this_thread::sleep_for(std::chrono::milliseconds(100));
```
GhostofCookie, line 1682: Should this sleep be configurable? Or is 100ms okay

src/transport_picoquic.cpp
```diff
@@ -1629,121 +1660,172 @@ PicoQuicTransport::CheckConnsForCongestion()
...
+        SPDLOG_LOGGER_ERROR(logger, "Failed to create picoquic network thread");
         picoquic_free(quic_ctx_);
-        quic_ctx_ = NULL;
+        quic_ctx_ = nullptr;
+        return false;
+    }
+
+    // Wait for something to happen with the thread
+    while (!quic_network_thread_ctx_->thread_is_ready && !quic_network_thread_ctx_->return_code) {
+        std::this_thread::sleep_for(std::chrono::milliseconds(100));
```
GhostofCookie, line 1817: Same question

GhostofCookie (dismissed):
This looks interesting! I'm not sure, but for every time we Push to the runner queue, do we need to call the wake up? I feel like I saw a few spots where we don't do it, so not sure if it's required or not

src/transport_picoquic.cpp
TimEvens: Yeah, I was thinking that too.  I'll update it.

src/transport_picoquic.cpp
TimEvens: Not all. The client one does return a picoquic error.  In the past using picoquic sync loop, we didn't need to use this as much. Now with the network thread, future changes will require us to use this more.  We might need those to exit in a similar way as start client.

src/transport_picoquic.cpp
```diff
@@ -1629,121 +1660,172 @@ PicoQuicTransport::CheckConnsForCongestion()
...
+    SPDLOG_LOGGER_DEBUG(logger, "Got client connection context conn_id: {}", conn_id);
+    if (conn_id <= 1) {
+        SPDLOG_LOGGER_DEBUG(logger, "Client connection to {}:{} failed", serverInfo_.host_or_ip, serverInfo_.port);
         SetStatus(TransportStatus::kDisconnected);
-        return;
+        return 0;
     }
 
-    SPDLOG_LOGGER_INFO(logger, "Thread client packet loop for client conn_id: {0}", conn_id);
+    picoquic_wake_up_network_thread(quic_network_thread_ctx_);
```
GhostofCookie, line 1785: Do we need this call to wake up thread given we just called it on line 1772?

src/transport_picoquic.cpp
TimEvens: It's only on start of the server, but 100ms might be annoying.  I've changed it to 1ms.

src/transport_picoquic.cpp
TimEvens: good catch, no. I"ll remove that.

GhostofCookie (commented):
Looks good, though tests are seg faulting on macos.

GhostofCookie (approved):

src/transport_picoquic.cpp
RichLogan: This is in config now, but still hardcoded here, is that intentional?

RichLogan (approved):

PR #703, opened 2025-09-30
Rework fetch.

### LINKED ISSUE #699: Refactor Fetch in libquicr and LAPS for single relay
Fetch doesn't follow the design of using `<message>Received()` callback followed by `<message>Resolved()` API method. 

- [ ] Remove fetch and cache from qserver. This is part of stripping down qserver in favor of using LAPS
- [ ] Remove `OnFetch()`
- [ ] `FetchReceived()` callback should be called and implemented similar to `SubscribeReceived()` . It doesn't need to return anything back to quicr::Server/quicr::Client
- [ ] Implementation of handling the fetch should be removed from `ProcessCtrMessage()` and moved to LAPS and the client (`qclient`)
- [ ] Implement `ResolveFetch()` similar to `ResolveSubscribe()`
- [ ] Add `OnStreamClosed()` callback to QUIC Transport to indicate if stream has been closed by FIN or RST.  Implement this in transport fetch to close/clear out the fetch receive track. Notify the fetch track handler via status update that it's `kDoneByFIN` or `kDoneByReset`. These two new status/states are specific to fetch and not subscribe track handler.   
- [ ] Implement a mini cache in `qclient` to cache last X (cmd arg) groups. This will be used by `qclient` publisher to fulfill the fetch request.
- [ ] Client should return cache in a similar way that the server does

In LAPS, the cache may need to be updated to handle the changes.  LAPS will need to handle the above changes to correctly implement fetch based on MOQT.  Fetch needs to work with both announce and publish flows. 

In **standalone** fetch only; If there is no cached objects or the range is not within the available cache, the fetch should be sent to the publisher, which may need to be matched on announce using prefix matching.  In the case of multiple matches, the longest namespace prefix match should be used (aka more specific or longest namespace match). If there are multiple publishers, only the first one will get the fetch request. Peering will be handled via a different issue/PR.

GhostofCookie:
Closes #699

### DIFF
--- a/src/transport_picoquic.cpp
+++ b/src/transport_picoquic.cpp
@@ -1677,7 +1677,7 @@ PicoQuicTransport::CheckConnsForCongestion()
 
         } else if (conn_ctx.is_congested) {
 
-            if (conn_ctx.not_congested_gauge > 4) {
+            if (conn_ctx.not_congested_gauge > 8) {
                 // No longer congested
                 conn_ctx.is_congested = false;
                 conn_ctx.not_congested_gauge = 0;


include/quicr/server.h
```diff
@@ -160,6 +160,24 @@ namespace quicr {
                                     messages::GroupOrder group_order,
                                     const PublishResponse& publish_response);
 
+        /**
+         * @brief Accept or reject a Fetch that was received
+         *
+         * @details Accept or reject an fetch received via FetchReceived().
```
RichLogan, line 166: an -> a

src/transport_picoquic.cpp
```diff
@@ -200,6 +202,8 @@ PqEventCb(picoquic_cnx_t* pq_cnx,
             SPDLOG_LOGGER_TRACE(
               transport->logger, "Received RESET stream conn_id: {0} stream_id: {1}", conn_id, stream_id);
 
+            transport->OnStreamClosed(conn_id, stream_id, true, false);
```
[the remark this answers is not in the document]
GhostofCookie: Oops! Copy paste error, good catch!

src/transport_picoquic.cpp
```diff
@@ -1302,9 +1306,12 @@ PicoQuicTransport::SendStreamBytes(DataContext* data_ctx, uint8_t* bytes_ctx, si
             data_ctx->tx_start_stream = false;
 
         } else {
-            // Queue is empty
             picoquic_provide_stream_data_buffer(bytes_ctx, 0, 0, not data_ctx->tx_data->Empty());
 
+            if (!data_ctx->tx_data->Empty() && data_ctx->delete_on_empty) {
```
GhostofCookie: oops!

PR #712, opened 2025-10-14
Add extensions benchmark

RichLogan:
Benchmark for extension serialise/deserialise.

### DIFF
--- a/benchmark/extensions.cpp
+++ b/benchmark/extensions.cpp
@@ -1,0 +1,150 @@
+#include <benchmark/benchmark.h>
+#include <quicr/detail/messages.h>
+#include <quicr/detail/stream_buffer.h>
+
+using namespace quicr;
+using namespace quicr::messages;
+
+// Helper function to create test extensions
+static Extensions
+CreateTestExtensions(const std::uint64_t count)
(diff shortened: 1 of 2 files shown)


TimEvens (approved):
LGTM

benchmark/extensions.cpp
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    state.SetBytesProcessed(static_cast<int64_t>(bytes_processed));
+    state.SetItemsProcessed(
+      static_cast<int64_t>(state.iterations() * (original_extensions.size() + original_immutable.size())));
+}
+
+// Benchmark round-trip (serialize + deserialize)
+static void
+ExtensionsRoundTrip(benchmark::State& state)
```
GhostofCookie: This is the benchmark that I would call the "real usage" that simulates how we actually use it.

GhostofCookie (approved):
This looks good, but I do think it's not benching what I would expect. We don't have any pure benchs for the serialisation and deserialisation. There is a lot of overlap in these 3 benchs, without revealing the real cost of the individual serialise and deserialise methods.

PR #715, opened 2025-10-14
Fixes for fetch and handling

TimEvens:
* Make transport close stream and data context on unpublish, unbindpublish and unbindfetch
* Update default cache duration
* Correct end-group setting in qclient
* Handle new status kDoneByFin and kDoneByReset
* Change various log messages to debug and trace
* Use RxStream track handler instead of for loop in OnStreamClosed


TimEvens:
Ignore the failing tests. It's an issue with GitHub where it is trying to run tests using the wrong base branch.  It's using main branch when it should be using `699_new-fetch`.  Normally GitHub would have automatically not run tests with this type of PR.

PR #724, opened 2025-10-22
Subscribe namespace publish flow

RichLogan:
Adds PUBLISH flow for an accepted subscribe namespace publish notification, causing data to flow by hooking into the subscribe flow on accept. 

The `is_publisher_initiated` flag will probably not need to exist once a new API is added specifically for multi-track handling - this is to allow the existing SUBSCRIBE flow and handlers to be used in lieu of a new handler type, while suppressing the OK flow.

### DIFF
--- a/src/server.cpp
+++ b/src/server.cpp
@@ -92,6 +92,10 @@ namespace quicr {
                                   uint64_t track_alias,
                                   const SubscribeResponse& subscribe_response)
     {
+        if (subscribe_response.is_publisher_initiated) {
+            throw std::invalid_argument("ResolveSubscribe should not be called when publisher initiated");
+        }
+
         auto conn_it = connections_.find(connection_handle);
         if (conn_it == connections_.end()) {
             return;
@@ -113,10 +117,8 @@ namespace quicr {
                 req_it->second.largest_location = subscribe_response.largest_location;
 
                 // Send the ok.
-                if (!subscribe_response.is_publisher_initiated) {
-                    SendSubscribeOk(
-                      conn_it->second, request_id, track_alias, kSubscribeExpires, subscribe_response.largest_location);
-                }
+                SendSubscribeOk(
+                  conn_it->second, request_id, track_alias, kSubscribeExpires, subscribe_response.largest_location);
                 break;
             }
             default:
--- a/cmd/examples/server.cpp
+++ b/cmd/examples/server.cpp
@@ -881,15 +881,17 @@ class MyServer : public quicr::Server
 
         const auto track_alias = th.track_fullname_hash;
 
-        ResolveSubscribe(connection_handle,
-                         request_id,
-                         track_alias,
-                         {
-                           quicr::SubscribeResponse::ReasonCode::kOk,
-                           attrs.is_publisher_initiated,
-                           std::nullopt,
-                           largest_location,
-                         });
+        if (!attrs.is_publisher_initiated) {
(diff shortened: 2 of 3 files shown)


src/client.cpp
```diff
@@ -31,6 +31,37 @@ namespace quicr {
     {
     }
 
+    void Client::PublishReceived(const ConnectionHandle,
+                                 const messages::TrackAlias,
+                                 const FullTrackName&,
+                                 const messages::RequestID)
+    {
```
suhasHere: do we need override tag here ?

src/server.cpp
```diff
@@ -967,14 +986,87 @@ namespace quicr {
...
 
-                PublishReceived(
-                  conn_ctx.connection_handle,
-                  msg.request_id,
-                  tfn,
-                  { { 0, msg.group_order, std::chrono::milliseconds(0), 0, new_group_request_id }, msg.track_alias });
+                PublishReceived(conn_ctx.connection_handle,
+                                msg.request_id,
+                                tfn,
+                                { { 0,
```
suhasHere: what is this 0 for

suhasHere (dismissed):
few nits. Otherwise LGTM.

please add issues to TODO so we don't forget it.

src/server.cpp
RichLogan: For posterity, this is here because publish attributes are a superset of subscribe attributes, which is not actually ideal for things like subscriber priority here. Will open an issue to resolve that relationship.

src/client.cpp
RichLogan: #729

include/quicr/subscribe_track_handler.h
```diff
@@ -225,6 +225,12 @@ namespace quicr {
          */
         void SupportNewGroupRequest(bool is_supported) noexcept;
 
+        /**
+         * @brief Indicate if subscribe handler can send new group requests or not
+         * @return True if new group requests are supported.
+         */
+        bool NewGroupRequestSupported() const noexcept { return support_new_group_request_; }
```
TimEvens: Very close naming to `SupportNewGroupRequest()`.  Maybe rename this to `IsNewGroupRequestSupported()` 

This also could rathole with MOQT adding more params that could cause us to have  `SetXSupported()` and `IsXSupported()` pattern for various features.

src/client.cpp
```diff
@@ -549,6 +580,29 @@ namespace quicr {
...
+                messages::Publish msg([](messages::Publish& msg) {
+                    if (msg.content_exists) {
+                        msg.group_0 =
+                          std::make_optional<messages::Publish::Group_0>(messages::Publish::Group_0{ 0, 0 });
+                    }
+                });
+                msg_bytes >> msg;
+
+                const FullTrackName tfn{ .name_space = msg.track_namespace, .name = msg.track_name };
+                const TrackHash th(tfn);
```
TimEvens: Params are not being parsed and passed to the app.  For example, dynamic groups param.

src/server.cpp
```diff
@@ -967,14 +986,87 @@ namespace quicr {
...
+                                    messages::FilterType::kLargestObject,
+                                    0,
+                                    new_group_request_id,
+                                    true },
+                                  msg.track_alias });
 
                 return true;
             }
+
+            case messages::ControlMessageType::kPublishOk: {
```
TimEvens: We should be able to consolidate some here with subscribe ok.

PR #740, opened 2025-11-13
Qbridge - C Wrapper around libQuicr

suhasHere:


c-bridge/include/quicr/quicr_bridge.h
```diff
@@ -0,0 +1,746 @@
...
+                                            qbridge_connection_status_callback_t callback,
+                                            void* user_data);
+
+    /**
+     * @brief Set namespace callback
+     * @param client Client handle
+     * @param callback Callback function
+     * @param user_data User data to pass to callback
+     */
+    void qbridge_client_set_namespace_callback(qbridge_client_t* client,
```
RichLogan: Not sure if there's a better name for this or not

PR #743, opened 2025-11-13
Fixes for subscribe namespace and renames

TimEvens:
This PR includes additional perf and other fixes as well as subscribe namespace updates and renames.

### DIFF
--- a/include/quicr/publish_track_handler.h
+++ b/include/quicr/publish_track_handler.h
@@ -356,8 +356,9 @@ namespace quicr {
          */
         void SetStatus(Status status) noexcept
         {
-            if (publish_status_ == status)
+            if (publish_status_ == status) {
                 return;
+            }
 
             publish_status_ = status;
             StatusChanged(status);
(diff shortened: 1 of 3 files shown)


src/transport_picoquic.cpp
```diff
@@ -515,9 +515,9 @@ PicoQuicTransport::Start()
     // picoquic_set_default_wifi_shadow_rtt(quic_ctx, tconfig.quic_wifi_shadow_rtt_us);
     // logger->info << "Setting wifi shadow RTT to " << tconfig.quic_wifi_shadow_rtt_us << "us" << std::flush;
 
-    picoquic_runner_queue_.SetLimit(2000);
+    picoquic_runner_queue_.SetLimit(10000);
```
suhasHere: just trying to understand. why did we increase these limits for this feature ?

src/transport_picoquic.cpp
```diff
@@ -515,9 +515,9 @@ PicoQuicTransport::Start()
...
 
-    cbNotifyQueue_.SetLimit(2000);
+    cbNotifyQueue_.SetLimit(10000);
```
suhasHere: same question as above

RichLogan (dismissed):
Not got any intel on the perf stuff, but protocol changes LGTM.

include/quicr/publish_track_handler.h
```diff
@@ -356,6 +356,9 @@ namespace quicr {
          */
         void SetStatus(Status status) noexcept
         {
+            if (publish_status_ == status)
```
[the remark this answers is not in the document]
TimEvens: Will wrap with braces.

RichLogan (approved):

PR #759, opened 2026-01-05
Make client consistent with server handling

RichLogan:
Makes the client inbound FETCH handling consistent with the existing server side FETCH handling. 

(This duplication can be removed, but for now just making them behave the same way).


TimEvens (approved):
LGTM

PR #764, opened 2026-01-21
Add Sframe MLS to QClient

GhostofCookie:
Add the ability to use sframe protect/unprotect on data in qclient.


cmd/examples/client.cpp
```diff
@@ -180,7 +189,12 @@ class MySubscribeTrackHandler : public quicr::SubscribeTrackHandler
 
     void ObjectReceived(const quicr::ObjectHeaders& hdr, quicr::BytesSpan data) override
     {
-        if (qclient_vars::record) {
+        quicr::Bytes pt(data.size());
```
RichLogan: Minor but I guess this only needs to be actually allocated with a real size in the block?

Issue #770, opened 2026-02-02
Data Messages changes to comply with draft-16


suhasHere:
Following messages needs revisiting 
- ObjectDatagram  (Type has been updated)
- SubgroupHeader (Type has been updated)
- Match the behavior of Closing subgroup streams
- Fetch headers ( bitmask changes)

PR #778, opened 2026-02-04
Datagram header bitfields

RichLogan:
Refactor datagram header types for draft-16, using the now explicit bitfields. ref #770. See also note about publisher priority which isn't implemented yet: #777

### DIFF
--- a/src/transport.cpp
+++ b/src/transport.cpp
@@ -1974,8 +1974,10 @@ namespace quicr {
                 auto msg_type = data->front();
 
                 // Message type needs to be either datagram header types or status types.
-                const auto data_type = static_cast<DatagramMessageType>(msg_type);
-                if (!TypeIsDatagram(data_type)) {
+                try {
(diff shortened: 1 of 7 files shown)


src/transport.cpp
```diff
@@ -1974,8 +1974,10 @@ namespace quicr {
                 auto msg_type = data->front();
 
                 // Message type needs to be either datagram header types or status types.
-                const auto data_type = static_cast<DatagramMessageType>(msg_type);
-                if (!TypeIsDatagram(data_type)) {
+                try {
```
RichLogan: Rich: Probably shouldn't use exception for control flow here. Can just fix this *but* are there other protocol violations that could bubble up here and be handled as a connection close at the `OnRecvDgram`. Need to read the code a bit more around this. Maybe for now I make non-throwing, to keep existing behaviour, and open an issue about protocol violation handling.

TimEvens (approved):
LGTM

PR #785, opened 2026-02-06
Support new fetch object serialization variants

RichLogan:
New fetch object serialization. It's getting late so I've thrown this up - I think it's almost there but haven't deeply reviewed what I've done. There are some assertions in here I need to check are actual assertions and not protocol violations. I have tried to keep things type safe / explicit where I can. Now that basically everything is optional, it's harder to easily reason about the valid states.


TimEvens:
Old, has been redone.

PR #796, opened 2026-03-02
Ensure object ID deltas calculated correctly

RichLogan:
On new streams, the `StreamInfo` was incorrectly being initialized with last object having a real value, when it should be null, to restart the delta encoding.


TimEvens (approved):
LGTM

PR #812, opened 2026-03-26
Fixes for interop and closing of streams on end of subgroup/FIN/RESET

TimEvens:
Fixes an issue causing Interop with openmoq/moxygen from working. 

This PR also fixes stream close by FIN/REST upon end of subgroup.

### DIFF
--- a/include/quicr/track_name.h
+++ b/include/quicr/track_name.h
@@ -11,7 +11,6 @@
 #include <span>
 #include <stdexcept>
 #include <string>
-#include <tuple>
 #include <vector>
 
 namespace quicr {
@@ -228,19 +227,21 @@ namespace quicr {
 
         friend bool operator>=(const TrackNamespace& lhs, const TrackNamespace& rhs) noexcept { return !(lhs < rhs); }
 
-        bool IsPrefixOf(const TrackNamespace& other) const noexcept
+        std::partial_ordering IsPrefixOf(const TrackNamespace& rhs) const noexcept
         {
-            if (this->hashes_.size() > other.hashes_.size()) {
-                return false;
+            if (!HasSamePrefix(rhs)) {
+                return std::partial_ordering::unordered;
             }
 
-            for (size_t i = 0; i < this->hashes_.size(); i++) {
-                if (this->hashes_[i] != other.hashes_[i]) {
-                    return false;
-                }
+            if (entries_.size() < rhs.entries_.size()) {
+                return std::partial_ordering::less;
             }
 
-            return true;
+            if (entries_.size() > rhs.entries_.size()) {
+                return std::partial_ordering::greater;
+            }
+
+            return std::partial_ordering::equivalent;
         }
 
         bool HasSamePrefix(const TrackNamespace& other) const noexcept
--- a/include/quicr/track_name.h
+++ b/include/quicr/track_name.h
@@ -9,6 +9,7 @@
 #include <cstdint>
 #include <optional>
 #include <span>
+#include <sstream>
 #include <stdexcept>
 #include <string>
 #include <vector>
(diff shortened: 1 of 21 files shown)


GhostofCookie (changes requested):

src/publish_track_handler.cpp
```diff
@@ -415,4 +415,18 @@ namespace quicr {
...
+        for (auto& [group_id, subgroups] : stream_info_by_group_) {
+            for (auto& [subgroup_id, stream_info] : subgroups) {
+                if (stream_info.stream_id == stream_id) {
+                    subgroups.erase(subgroup_id);
+                    if (subgroups.empty()) {
+                        stream_info_by_group_.erase(group_id);
+                    }
+                    return;
+                }
+            }
```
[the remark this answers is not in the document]
suhasHere: how about we do 
```
auto it = std::find_if(subgroups.begin(), subgroups.end(),
      [stream_id](const auto& p) { return p.second.stream_id ==
  stream_id; });

  if (it != subgroups.end()) {
      subgroups.erase(it);
      if (subgroups.empty()) {
          stream_info_by_group_.erase(group_id);
      }
  }
  ```

File: src/transport_picoquic.cpp [hunk 2]
```diff
@@ -896,9 +907,11 @@ PicoQuicTransport::Start()
 {
     uint64_t current_time = picoquic_current_time();
 
+#if 0
     if (debug) {
         debug_set_stream(stderr);
     }
+#endif
```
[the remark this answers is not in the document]
TimEvens: The change to use the latest picoquic introduced some noise with debug messages. We should upstream a change to those as they are likely not needed.  When that's fixed, we should add back.

src/publish_track_handler.cpp
```diff
@@ -82,9 +82,16 @@ namespace quicr {
                 eflags.use_reliable = true;
 
                 if (is_new_stream) {
+                    // Disable creating stream and doing pipeline on new stream considering delta and subgroups needs
+                    // first object parsed
+#if 0
```
GhostofCookie, line 87: Is the plan to finish implementing this before the PR is merged?

src/publish_track_handler.cpp
```diff
@@ -386,7 +380,7 @@ namespace quicr {
         transport->Enqueue(GetConnectionId(),
                            publish_data_ctx_id_,
                            subgroup_it->second.stream_id,
-                           std::make_shared<std::vector<uint8_t>>(object_msg_buffer_.begin(), object_msg_buffer_.end()),
+                           {},
```
GhostofCookie, line 385: Is this intentional? Or correct?

src/server.cpp
```diff
@@ -181,7 +181,8 @@ namespace quicr {
 
         // Fan out PUBLISH_NAMESPACE for matching namespaces.
         for (const auto& name_space : response.namespaces) {
-            if (!prefix.IsPrefixOf(name_space)) {
+            const auto match = prefix.IsPrefixOf(name_space);
+            if (match == std::partial_ordering::unordered || match == std::partial_ordering::less) {
```
GhostofCookie, line 185: Less would mean that `prefix` is the prefix of `name_space`. Greater would mean `name_space` is the prefix of `prefix`.

include/quicr/track_name.h
```diff
@@ -228,19 +231,21 @@ namespace quicr {
 
         friend bool operator>=(const TrackNamespace& lhs, const TrackNamespace& rhs) noexcept { return !(lhs < rhs); }
 
-        bool IsPrefixOf(const TrackNamespace& other) const noexcept
+        std::partial_ordering IsPrefixOf(const TrackNamespace& rhs) const noexcept
```
GhostofCookie, line 247: Maybe this should instead be a `operator<=>`, that way instead of calling things like
```cpp
auto match = prefix.IsPrefixOf(name_space)
if (math == std::partial_ordering::unordered || match == std::partial_ordering::less)
```
where `name_space` might actually be the prefix of `prefix`, instead we could use
```cpp
if (prefix != name_space || prefix < name_space)
```
or 
```cpp
if (!(prefix >= name_space))

GhostofCookie (approved):

PR #813, opened 2026-03-26
Dynamic groups for publisher initiated

RichLogan:
Fix client side dynamic groups subscribe handler always being 0 on publish initiated flow.


TimEvens (approved):
LGTM

PR #814, opened 2026-04-09
Rework group order for interop

RichLogan:
Group order was not quite right, and I figured removing the original order and modelling it as an optional matches the data model in the spec better, otherwise we might encode 0x0 on the wire as a group order, which is invalid. Also added the publisher side track extension API. There were a couple of places we were looking for group order where it didn't make sense, and those attribute structures still let you model invalid states, but I think this is cleaner. 

This also adds the publishers group order to use the proper track extension.


suhasHere (commented):
this looks good. 

Can we also add unit tests with zero, 3+ track extensions for a couple of control messages ( SUB, SUBOK, PUB, PUBOK)

TimEvens (approved):
LGTM

PR #815, opened 2026-04-09
Structured field encoding for WT protocols

RichLogan:
I believe that this field should be passed to picoquic as a structured field, and that wt_available_protocols is specifically a LIST of STRING fields, which means each item (of which we have 1) must be enclosed in double quotes.

ref: https://datatracker.ietf.org/doc/html/rfc8941 

This made moxygen WT interop start working for me.

### DIFF
--- a/src/transport_picoquic.cpp
+++ b/src/transport_picoquic.cpp
@@ -2437,6 +2437,7 @@ PicoQuicTransport::StartClient()
                                wt_config_->path);
 
             // Initiate the WebTransport connect
+            const std::string wt_available_protocols = std::string("\"") + moqt_alpn + "\"";
             ret = picowt_connect(cnx,
                                  h3_ctx,
                                  control_stream_ctx,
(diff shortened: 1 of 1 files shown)


suhasHere:
I understand this is needed. Let's not merge until I do few more other tests . Thanks

TimEvens (approved):
LGTM

src/transport_picoquic.cpp
```diff
@@ -2437,14 +2437,15 @@ PicoQuicTransport::StartClient()
                                wt_config_->path);
 
             // Initiate the WebTransport connect
+            const std::string wt_available_protocols = std::string("\"") + moqt_alpn + "\"";
```
TimEvens, line 2440: I believe PQ was adding the double quotes, but since has removed it.  IMO, it should be PQ that does this.

TimEvens:
close for now.

PR #816, opened 2026-04-09
Properly handle multi-message callbacks

RichLogan:
We should only process the declared size of the ctrl message, not till the end of the buffer, or we might starting reading bytes of the next control message into the previous one. I think this is showing up now because of track extensions? Since they don't have a length, if we pass too many bytes and try and parse a control message with track extensions, we'll interpret any additional message in the buffer as track extension bytes of the first one.

### DIFF
--- a/src/transport.cpp
+++ b/src/transport.cpp
@@ -1586,11 +1586,11 @@ namespace quicr {
                         break; // Not enough bytes to process control message. Try again once more.
                     }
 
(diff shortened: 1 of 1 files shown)


GhostofCookie (approved):

TimEvens (approved):
That's true

TimEvens (approved):
That's true

PR #818, opened 2026-04-14
Set PATH and AUTHORITY on raw QUIC client setup

RichLogan:


src/transport.cpp
```diff
@@ -350,6 +350,14 @@ namespace quicr {
                                   .Add(SetupParameterType::kEndpointId, client_config_.endpoint_id)
                                   .Add(SetupParameterType::kMaxRequestId, 0x1000);
 
+        // Set PATH and AUTHORITY for native QUIC connections.
+        // Start() has already validated parse of connect URI succeeds.
```
GhostofCookie, line 354: Not sure the comment is particularly necessary.

GhostofCookie (dismissed):

src/transport.cpp
RichLogan: I was trying to explain why that unchecked optional unwrap is safe in practise.

src/transport.cpp
```diff
@@ -350,6 +350,14 @@ namespace quicr {
...
+        const auto [authority, _, protocol, path] = *ParseConnectUri(client_config_.connect_uri);
+        if (protocol != TransportProtocol::kWebTransport) {
+            setup_parameters.Add(SetupParameterType::kAuthority, authority);
+            setup_parameters.Add(SetupParameterType::kPath, path);
+        }
+
```
suhasHere, line 363: this was my local change

```
auto [address, port, protocol, path] = *parse_result;
            if (protocol == TransportProtocol::kQuic) {
                if (!path.empty()) {
                    setup_parameters.Add(SetupParameterType::kPath, path);
                }

                std::string authority = address;
                if (port != 0 && port != 443) {
                    authority += ":" + std::to_string(port);
                }
                setup_parameters.Add(SetupParameterType::kAuthority, authority);

                SPDLOG_LOGGER_INFO(
                  logger_,
                  "ClientSetup native QUIC params: authority='{}' path='{}'",
                  authority,
                  path);
            }
```

RichLogan: I may well be wrong but In my reading of the draft `PATH` wasn't optional or omitted when empty. 

> the client MUST set the PATH option to the path-abempty portion of the URI

In fact, if it's empty, should it be `/` instead?

suhasHere: empty path is considered as just "/"

GhostofCookie (approved):

suhasHere (approved):

PR #820, opened 2026-04-16
Guard filter param

RichLogan:
`GetFilterParameterType()` can return `kInvalid` when there's no filter. The caller MUST check this and not send `kInvalid` as a parameter on the wire if that is the case. SUBSCRIBE was doing this correctly, but SUBSCRIBE_NAMESPACE and PUBLISH_OK were not. 

Possibly filters could be optional, return nullopt on no filter, and throw on an invalid, maybe that might reduce the chance of this happening, but probably not.

This fixes an interop bug against OpenMoQ.


GhostofCookie (approved):

GhostofCookie:
Might even be better to address it in the Parameters type, in the Add method just ignore any filter given if the type is invalid

PR #824, opened 2026-04-24
New message formatter

GhostofCookie:
Currently we generate messages via the draft parser, which creates a lot of classes and stream operators, and gets a bit awkward when it comes to upgrading. This new message formatter (currently only for writing) is a step towards reducing the message classes, and should make updating message formats in future draft updates much simpler.


include/quicr/detail/message.h
```diff
@@ -0,0 +1,80 @@
...
+      public:
+        template<typename T>
+            requires std::is_enum_v<T>
+        Message(T type)
+        {
+            auto type_bytes = UintVar(static_cast<std::uint64_t>(type));
+            _bytes.insert(_bytes.end(), type_bytes.begin(), type_bytes.end());
+        }
+
+        Message& ReserveLength()
```
TimEvens: Why can't this be part of the constructor when adding type?  Having it as a separate call could introduce errors, for example, call `Add(...)` then call `ReserveLength()`.

include/quicr/detail/message.h
```diff
@@ -0,0 +1,80 @@
...
+#include "ctrl_message_types.h"
+#include "uintvar.h"
+#include "utilities.h"
+
+#include <cstdint>
+#include <span>
+#include <vector>
+
+namespace quicr::messages {
+    class Message
```
TimEvens, line 13: This really should be renamed to `ControlMessage` since this class is just for control message handling.

include/quicr/detail/message.h
```diff
@@ -0,0 +1,80 @@
...
+            }
+
+            const std::size_t type_size = UintVar(_bytes.front()).size();
+            _bytes.insert(std::next(_bytes.begin(), type_size), { 0, 0 });
+
+            _length_reserved = true;
+            return *this;
+        }
+
+        Message& Add(std::uint8_t byte) noexcept
```
TimEvens: Maybe more intuitive as `Append(...)`?

include/quicr/detail/message.h
```diff
@@ -0,0 +1,80 @@
...
+            return _bytes;
+        }
+
+      private:
+        bool _length_reserved = false;
+        mutable std::vector<std::uint8_t> _bytes;
+    };
+
+    template<typename Field>
+    static Field ParseMessageField(std::span<const std::uint8_t>& buffer)
```
TimEvens: Does this compile to be inline?

include/quicr/detail/transport.h
```diff
@@ -546,6 +547,20 @@ namespace quicr {
...
 
         void SendCtrlMsg(const ConnectionContext& conn_ctx, DataContextId data_ctx_id, BytesSpan data);
+
+        template<typename... Fields>
+        void SendCtrlMsg(const ConnectionContext& conn_ctx,
+                         DataContextId data_ctx_id,
+                         messages::ControlMessageType type,
+                         Fields&&... args)
+        {
+            auto msg = messages::Message{ type }.ReserveLength();
```
TimEvens: This is what's a bit weird to call `ReserveLength()`.  IMHO, it would be more intuitive to construct the type and move past the control message length upon construct.

include/quicr/detail/message.h
```diff
@@ -0,0 +1,80 @@
...
+        const std::vector<std::uint8_t>& ToBytes() const noexcept
+        {
+            if (_length_reserved) {
+                const std::size_t type_size = UintVar(_bytes.front()).size();
+                const std::uint16_t payload_size =
+                  SwapBytes(static_cast<std::uint16_t>(_bytes.size() - type_size - sizeof(std::uint16_t)));
+                std::memcpy(_bytes.data() + type_size, &payload_size, sizeof(std::uint16_t));
+            }
+
+            return _bytes;
```
TimEvens, line 78: Would returning a span be more direct and hopefully avoid copies?

TimEvens (commented):
Nice change and looks good, minor comments.

include/quicr/detail/transport.h
GhostofCookie: The idea is that you could also build messages that don't have a reserved length, like the data messages.

include/quicr/detail/message.h
GhostofCookie: Not necessarily. I only added it for control messages just now, but in theory you can use if for the data messages too.

include/quicr/detail/message.h
GhostofCookie: It would avoid copies entirely, but at the cost of building a span (which is minimal). I figured returning a const ref was good enough but maybe span is better.

include/quicr/detail/message.h
GhostofCookie: Should do since it's a template, but we can help the compiler a bit more.

include/quicr/detail/message.h
TimEvens: Problem is that data message does not have 2 byte length.

include/quicr/detail/transport.h
TimEvens: Makes sense to then change `ReserveLength()` to a constructor parameter that indicates if the message is control or not.

TimEvens (approved):
Good per our discussion.

PR #825, opened 2026-04-24
Fix init order to stop error on warning about init ordering.

GhostofCookie:


BrettRegnier (approved):
Fixes Werror=reorder errors on esp32s3

PR #826, opened 2026-05-12
Improve Build Systems

GhostofCookie:
- General clean up of cmake files, especially the root one for easier understanding at a glance and ensuring our build options are set correctly.
- Removed most submodules and replaced them with CPM (Cmake Package Manager).
- Suppressed warnings for dependencies to clean up output to only show warning and errors for our code.
- Fixed CI to no longer use cache and allow for re-running tests and benchmarks independently without needing to rebuild the whole pipeline.


src/CMakeLists.txt
```diff
@@ -6,39 +6,35 @@ if (${QUICR_BUILD_SHARED})
...
+    subscribe_namespace_handler.cpp
+    subscribe_track_handler.cpp
     transport_picoquic.cpp
-    joining_fetch_handler.cpp
+    transport.cpp
 )
 
+target_compile_options(quicr PRIVATE
+    $<$<OR:$<CXX_COMPILER_ID:Clang>,$<CXX_COMPILER_ID:AppleClang>,$<CXX_COMPILER_ID:GNU>>: -Wpedantic -Wextra -Wall -Werror=switch>
+    $<$<CXX_COMPILER_ID:MSVC>: >
```
GhostofCookie: Should probably remove this MSVC reference, since we don't support windows anyway.

src/CMakeLists.txt
```diff
@@ -49,23 +45,18 @@ if(MSVC)
...
 
-target_link_libraries(quicr
-    PUBLIC
-        spdlog)
+target_link_libraries(quicr PUBLIC spdlog)
 
-target_link_libraries(quicr
-    PUBLIC
-        picoquic-core picoquic-log picohttp-core)
+target_link_libraries(quicr PUBLIC picoquic-core picoquic-log picohttp-core)
```
GhostofCookie: This should probably be private

src/CMakeLists.txt
```diff
@@ -49,23 +45,18 @@ if(MSVC)
...
-
 if(LINT)
     include(Lint)
     lint(quicr)
 endif()
 
-target_link_libraries(quicr
-    PUBLIC
-        spdlog)
+target_link_libraries(quicr PUBLIC spdlog)
```
GhostofCookie: This would ideally be a private dependency of quicr that we don't expose to other libraries (even though LAPS depends on this, LAPS should include it itself and not rely on it coming from quicr).

RichLogan (dismissed):
LGTM (on the assumption we're happy with CPM).

test/CMakeLists.txt
```diff
@@ -1,52 +1,45 @@
...
-    target_compile_options(quicr_test PRIVATE  -Wall -pedantic -Wextra)
-endif()
+target_compile_options(quicr_test PRIVATE
+    $<$<OR:$<CXX_COMPILER_ID:Clang>,$<CXX_COMPILER_ID:AppleClang>,$<CXX_COMPILER_ID:GNU>>: -Wpedantic -Wextra -Wall>
+)
 
-if(MSVC)
-    target_compile_options(quicr_test PRIVATE /W4 /WX)
-    target_compile_definitions(quicr_test _CRT_SECURE_NO_WARNINGS)
+if (doctest_ADDED)
```
RichLogan: I had an issue with this not working when CPM resolves a locally installed doctest that I happened to have through homebrew. In that case I changed this to:

```diff
 if (doctest_ADDED)
     list(APPEND CMAKE_MODULE_PATH "${doctest_SOURCE_DIR}/scripts/cmake")
-    include(doctest)
+elseif (doctest_DIR)
+    list(APPEND CMAKE_MODULE_PATH "${doctest_DIR}")
 endif()
+include(doctest)
```

RichLogan (approved):

PR #827, opened 2026-05-13
Improve Message

GhostofCookie:
Cleans up the message builder to not require the order of type or length be correctly called for them to be inserted at the correct positions.

### DIFF
--- a/include/quicr/detail/message.h
+++ b/include/quicr/detail/message.h
@@ -49,6 +49,11 @@ namespace quicr::messages {
         {
             UintVar type_bytes(static_cast<std::uint64_t>(type));
             _bytes->insert(_bytes->begin(), type_bytes.begin(), type_bytes.end());
+
+            if (_length_bytes_offset.has_value()) {
+                _length_bytes_offset = type_bytes.size();
+            }
+
             return *this;
         }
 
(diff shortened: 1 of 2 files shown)


include/quicr/detail/message.h
```diff
@@ -21,28 +27,54 @@ namespace quicr::messages {
...
+            if (_length_bytes_offset.has_value()) {
                 return *this;
             }
 
-            const std::size_t type_size = UintVar(_bytes->front()).size();
-            _bytes->insert(std::next(_bytes->begin(), type_size), { 0, 0 });
+            if (_bytes->empty()) {
+                _length_bytes_offset = 0;
+            } else {
+                _length_bytes_offset = UintVar::Size(_bytes->front());
```
RichLogan, line 73: I wonder if this should only happen if type has been prepended, and we set a flag? Tiny bit more state but you'll be off in bad territory if you wrote some payload bytes, *then* reserved the length, and it treated the first byte as a uintvar. 

I know we can't protect everything, but this one seems coverable. 

Or just kick off if you've written a payload before doing this?

GhostofCookie: At what point do we consider it overengineered? We could track the state but, I'm not sure it's worth the extra code to save developers from themselves. It would be odd to do work before reserving the space for the length. 

Maybe just blowing it up is best, since babying the developers isn't really going to teach them anything.

RichLogan (approved):

PR #828, opened 2026-05-18
Improve Time Queue

GhostofCookie:


TimEvens (approved):
LGTM... tests need to be passing though.

RichLogan:
Think that's combo of the buggy picoquic test and buggy GitHub action not correct re-running if you only re-run failed. I think they should either do single step build & test, or use artefacts instead of caching, otherwise choosing to only re-run failed will never pass.

PR #830, opened 2026-05-20
Stop status reset overwriting concurrent update

RichLogan:
Sometimes we "consume" a temporary status like `kNewGroupRequested` - reverting the status to `kOk` on the app's thread when they call `PublishObject()`. However, this can overwrite a remote status update that is set from the libquicr thread, while `PublishObject` is running. This change ensures we will only flip the status to `kOk` when consuming a temporary status when the current handler status is still said temporary status at the point of flipping it. 

e.g previously, the following sequence could occur. 

1. New Group Requested
2. Publish Object
3. Remote status event changes to some other status mid publish object call.
4. Publish Object call flips to `kOk` assuming it's still New Group Requested. 

Now 4 will not happen, because the strong exchange will not flip to kOk if the current status is not the New Group Requested status we are consuming.

Note that the object that was proposed to be published with a valid status will still be published even if the status has been changed to a non-publishing one in the meantime. This is an open question, but it's the existing behaviour and I didn't change it. fwiw I don't think we need the strictness and the complexity that would come with it(?)

### DIFF
--- a/src/publish_track_handler.cpp
+++ b/src/publish_track_handler.cpp
@@ -46,7 +46,7 @@ namespace quicr {
                 return PublishObjectStatus::kNotAuthorized;
             case Status::kNewGroupRequested:
                 [[fallthrough]];
-            case Status::kSubscriptionUpdated:
+            case Status::kSubscriptionUpdated: {
                 // reset the status to ok to imply change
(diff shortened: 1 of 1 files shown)


RichLogan:
I'm gonna reopen this.

PR #838, opened 2026-05-23
Performance improvements and time queue move to submodule

TimEvens:
Most of the change involves time queue fixes, which now has moved to github.com/quicr/timeq.  This PR updates libquicr to use the new timeq repo. 

perf updates

- remove gap checking in publish handler
- always run pq runner in pq loop callback
- wake only if runner queue is empty
- use select instead of poll with pq


include/quicr/detail/quic_transport.h
```diff
@@ -93,7 +93,7 @@ namespace quicr {
         std::string tls_key_filename;                ///< QUIC TLS private key to use
         uint32_t time_queue_init_queue_size{ 1000 }; ///< Initial queue size to reserve upfront
         uint32_t time_queue_max_duration{ 2000 };    ///< Max duration for the time queue in milliseconds
-        uint32_t time_queue_bucket_interval{ 1 };    ///< The bucket interval in milliseconds
+        uint32_t time_queue_bucket_interval{ 500 };  ///< The bucket interval in milliseconds
```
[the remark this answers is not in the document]
TimEvens: this has been fixed in the timeq repo

dependencies/CMakeLists.txt
```diff
@@ -140,6 +143,7 @@ else ()
 
     set(PICOQUIC_FETCH_PTLS ON)
     set(picoquic_BUILD_TESTS OFF)
+    set(WITH_SELECT OFF)
```
GhostofCookie: Maybe we should remove this and start setting it via the command line when we configure the build folder? That way we can toggle it faster.

dependencies/CMakeLists.txt
```diff
@@ -29,6 +29,10 @@ CPMAddPackage("gh:gabime/spdlog@1.17.0")
 set(BUILD_SHARED_LIBS OFF)
 set(BUILD_STATIC_LIBS ON)
 
+# Time queue
+set(timeq_BUILD_BENCHMARKS OFF)
```
GhostofCookie: This should be turned off by default on the latest version of timeq.

include/quicr/detail/priority_queue.h
```diff
@@ -106,74 +110,66 @@ namespace quicr {
...
-
             std::map<uint8_t, std::vector<uint64_t>> remove_group_ids;
 
-            std::lock_guard _(mutex_);
             defer(RemoveGroupTimeQueue(remove_group_ids));
 
             for (auto& [priority, queue] : queues_) {
                 for (auto& [group_id, tqueue] : queue) {
-                    tqueue->PopFront(elem);
+                    auto elem = tqueue->pop_front();
```
GhostofCookie, line 159:
```suggestion
                    const auto [value, expired] = tqueue->pop_front();
```

test/priority_queue.cpp
```diff
@@ -7,28 +7,29 @@
...
     }
 
-    TimeQueueElement<std::vector<uint8_t>> elem;
-
     for (size_t i = 0; i < 500; ++i) {
         CHECK(pq.Empty() == false);
-        pq.PopFront(elem);
-        CHECK(elem.has_value == true);
-        CHECK(elem.expired_count == 0);
+        auto elem = pq.PopFront();
```
GhostofCookie:
```suggestion
        const auto [value, expired] = pq.PopFront();
```

test/time_queue.cpp
GhostofCookie, line 1: Should this file be deleted? These tests exist in the timeq repo.

src/transport_picoquic.cpp
```diff
@@ -1262,6 +1269,34 @@ PicoQuicTransport::CloseInternal(const TransportConnId& conn_id, uint64_t app_re
     if (conn_it == conn_context_.end())
         return;
 
+    // Clear all stream TX queues and RX buffers to release shared pointers
+    for (auto& [data_ctx_id, data_ctx] : conn_it->second.active_data_contexts) {
+        for (auto& [stream_id, stream_ctx] : data_ctx.streams) {
+            if (stream_ctx.tx_data) {
+                stream_ctx.tx_data->Clear();
+                stream_ctx.tx_data.reset();
```
GhostofCookie: If we are resetting the whole pointer then calling clear is kind of redundant, no?

src/transport_picoquic.cpp
```diff
@@ -1262,6 +1269,34 @@ PicoQuicTransport::CloseInternal(const TransportConnId& conn_id, uint64_t app_re
...
+    for (auto& [stream_id, rx_buf] : conn_it->second.rx_stream_buffer) {
+        if (rx_buf.rx_ctx) {
+            rx_buf.rx_ctx->data_queue.Clear();
+        }
+    }
+
+    // Clear datagram RX and TX queues and reset shared pointers
+    if (conn_it->second.dgram_rx_data) {
+        conn_it->second.dgram_rx_data->Clear();
+        conn_it->second.dgram_rx_data.reset();
```
GhostofCookie: Same as above

src/transport_picoquic.cpp
```diff
@@ -1262,6 +1269,34 @@ PicoQuicTransport::CloseInternal(const TransportConnId& conn_id, uint64_t app_re
...
+    }
+    if (conn_it->second.dgram_tx_data) {
+        conn_it->second.dgram_tx_data->Clear();
+        conn_it->second.dgram_tx_data.reset();
```
GhostofCookie: Same as above

src/transport_picoquic.cpp
```diff
@@ -1631,20 +1665,21 @@ PicoQuicTransport::SendNextDatagram(ConnectionContext* conn_ctx, uint8_t* bytes_
         }
     };
 
-    TimeQueueElement<ConnData> out_data;
-    conn_ctx->dgram_tx_data->Front(out_data);
-    if (out_data.has_value) {
-        const auto data_ctx_it = conn_ctx->active_data_contexts.find(out_data.value.data_ctx_id);
+    std::lock_guard _(*conn_ctx->dgram_tx_data);
+
+    auto out_data = conn_ctx->dgram_tx_data->Front();
```
GhostofCookie:
```suggestion
    const auto [out_data, expired] = conn_ctx->dgram_tx_data->Front();
```

src/transport_picoquic.cpp
```diff
@@ -1718,15 +1763,16 @@ PicoQuicTransport::SendStreamBytes(DataContext* data_ctx, std::uint64_t stream_i
...
     CheckCallbackDelta(data_ctx);
 
-    TimeQueueElement<ConnData> obj;
+    std::lock_guard _(*stream_ctx.tx_data);
 
     if (data_ctx != nullptr && stream_ctx.tx_reset_wait_discard) { // Drop TX objects till next reset/new stream
-        stream_ctx.tx_data->Front(obj);
-        if (obj.has_value && obj.value.data) {
+
+        auto obj = stream_ctx.tx_data->Front();
```
GhostofCookie:
```suggestion
        const auto [value, _] = stream_ctx.tx_data->Front();
```

src/transport_picoquic.cpp
```diff
@@ -2862,6 +2913,7 @@ PicoQuicTransport::MarkStreamActive(const TransportConnId conn_id,
         stream_ctx = &data_ctx_it->second;
     }
 
+    picoquic_mark_active_stream(conn_it->second.pq_cnx, stream_id, 0, NULL);
```
GhostofCookie, line 2927: Should we leave a comment for future us to remind us why we did this hack?

GhostofCookie (dismissed):

src/transport_picoquic.cpp
RichLogan: If we do clear I think we need to lock tx_data here.

src/transport_picoquic.cpp
```diff
@@ -1262,6 +1269,34 @@ PicoQuicTransport::CloseInternal(const TransportConnId& conn_id, uint64_t app_re
...
+        }
+    }
+
+    // Clear datagram RX and TX queues and reset shared pointers
+    if (conn_it->second.dgram_rx_data) {
+        conn_it->second.dgram_rx_data->Clear();
+        conn_it->second.dgram_rx_data.reset();
+    }
+    if (conn_it->second.dgram_tx_data) {
+        conn_it->second.dgram_tx_data->Clear();
```
RichLogan: Needs to lock the queue I think?

src/transport_picoquic.cpp
```diff
@@ -2224,10 +2269,11 @@ PicoQuicTransport::CheckConnsForCongestion()
                 if (stream.tx_data->Size() >= 50) {
```
RichLogan, line 2278: Does this tx queue need a lock for Size() now?

src/transport_picoquic.cpp
RichLogan, line 2275: Here too? Or just extract it out once.

RichLogan, line 2663: Lock tx_data here too?

dependencies/CMakeLists.txt
TimEvens: removed

dependencies/CMakeLists.txt
TimEvens: removed.

include/quicr/detail/priority_queue.h
TimEvens: we return elem.

test/priority_queue.cpp
TimEvens: updated

src/transport_picoquic.cpp
TimEvens: Removed clear() and added lock.

src/transport_picoquic.cpp
TimEvens: removed clear()

src/transport_picoquic.cpp
TimEvens: updated.

src/transport_picoquic.cpp
TimEvens: Added.

src/transport_picoquic.cpp
TimEvens: Added.

src/transport_picoquic.cpp
TimEvens: Yes, added.

src/transport_picoquic.cpp
TimEvens: just once.

TimEvens: Added.

src/transport_picoquic.cpp
TimEvens: updated

TimEvens:
@RichLogan , @GhostofCookie 

I removed the tx_data.reset() because that was destroying while lock was being held.   That was there just in case and shouldn't be required considering the context is being destructed.

RichLogan (approved):

PR #839, opened 2026-05-27
Improve Time Queue Cleanup

GhostofCookie:
In the case that we are not popping data from the time queue, only pushing, then when we cleanup buckets, the space allocated is preserved, and so memory is never truly reduced.

Therefore there are 2 fixes in here:
1. Instead of calling std::vector::clear on the bucket, we actually erase the bucket entirely, freeing the memory.
2. Adding an optional future insertion index in Advance step to check if we are about to push to the queue after this Advance step, and if so check if that index wraps around to a bucket we just cleared/erased, and instead of erasing it, we keep the clear mechanic to avoid a deletion and reallocation in the insertion step.


GhostofCookie:
@codex review

TimEvens (approved):
LGTM and appears to be working well.

PR #841, opened 2026-06-02
Add ASAN and UBSAN options

RichLogan:
Add support for ASAN and UBSAN, fixes a possible use-after-free on fast shutdown (network thread ptr should go to nullptr) and a leak on shutdown (`picoquic_free`).

These attribute structs are going to go soon I'm sure, so these default values aren't really that important.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -3,6 +3,10 @@
 
 CPMAddPackage("gh:doctest/doctest@2.5.2")
 
+if(QUICR_ENABLE_SANITIZERS)
+    set(QUICR_SANITIZER_TEST_SRC "${CMAKE_CURRENT_SOURCE_DIR}/sanitizer_options.cpp")
+endif()
+
 add_executable(quicr_test
     cache.cpp
     client.cpp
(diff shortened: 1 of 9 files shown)


test/CMakeLists.txt
```diff
@@ -3,6 +3,10 @@
 
 CPMAddPackage("gh:doctest/doctest@2.5.2")
 
+if(QUICR_ENABLE_SANITIZERS)
+    set(QUICR_SANITIZER_TEST_SRC "${CMAKE_CURRENT_SOURCE_DIR}/sanitizer_options.cpp")
+endif()
+
```
GhostofCookie: Could instead use
```cmake
target_sources(quicr_test sanitizer_options.cpp)
```

test/moq_data_messages.cpp
```diff
@@ -517,6 +517,7 @@ StreamPerSubGroupObjectEncodeDecode(StreamHeaderProperties properties, Extension
         }
         // got one object
         object_count++;
+        std::destroy_at(&obj_out);
         std::construct_at(&obj_out);
```
GhostofCookie: Woah, what am I looking at here. This is definitely ***not*** how this was intended to be used. Why are we using destroy_at and construct_at to reinit this variable? That's absolutely wild. It would be safer to use
```suggestion
    obj_out = {};
```

RichLogan: Yeah this is a hack, we can't just assign because it's immutable / no assignment operator. We could use an optional and emplace though, that's nicer. I'll update.

TimEvens (approved):
LGTM

PR #843, opened 2026-06-02
Draft 18 varint

RichLogan:
Adds draft 18 varint. Because we can now handle a full 64 bit varint, some things go away (like exceptions in the value ctor, hashes.

Benchmarks look to be either ~the same or better.

Added tests for the test vectors that are now in the draft.


TimEvens (approved):
Nice.

PR #844, opened 2026-06-02
Collapse client server to Session

GhostofCookie:


RichLogan:
Merged into `draft-18`

PR #846, opened 2026-06-04
Compile thread safety

RichLogan:
Adds compile time lock checking for the thread safe queues. All violations were resolved in #838 but there is one false-positive to mark (because of `defer`). Moved the macOS job to use the already installed LLVM clang, since the built-in is too old for `-Wthread-safety`.

### DIFF
--- a/include/quicr/defer.h
+++ b/include/quicr/defer.h
@@ -7,10 +7,19 @@
 
 #define DEFER_CONCAT(a, b) DEFER_CONCAT_INNER(a, b)
 #define DEFER_CONCAT_INNER(a, b) a##b
+
+#if defined(__clang__)
+#define DEFER_PUSH _Pragma("clang diagnostic push") _Pragma("clang diagnostic ignored \"-Wthread-safety-analysis\"")
+#define DEFER_POP _Pragma("clang diagnostic pop")
+#else
+#define DEFER_PUSH
+#define DEFER_POP
+#endif
+
 #define defer(n)                                                                                                       \
     quicr::DeferType DEFER_CONCAT(defer_, __LINE__)([&] {                                                              \
-        _Pragma("clang diagnostic push") _Pragma("clang diagnostic ignored \"-Wthread-safety-analysis\"") n;           \
-        _Pragma("clang diagnostic pop")                                                                                \
+        DEFER_PUSH n;                                                                                                  \
+        DEFER_POP                                                                                                      \
     })
 
 namespace quicr {
--- a/include/quicr/defer.h
+++ b/include/quicr/defer.h
@@ -7,7 +7,11 @@
 
 #define DEFER_CONCAT(a, b) DEFER_CONCAT_INNER(a, b)
 #define DEFER_CONCAT_INNER(a, b) a##b
-#define defer(n) quicr::DeferType DEFER_CONCAT(defer_, __LINE__)([&] { n; })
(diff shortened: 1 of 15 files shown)


include/quicr/detail/priority_queue.h
```diff
@@ -3,11 +3,14 @@
 
 #pragma once
 
+#include <quicr/detail/thread_safety.h>
```
GhostofCookie: Prefer quotations for local headers
```suggestion
#include "quicr/detail/thread_safety.h"
```

src/transport_picoquic.cpp
```diff
@@ -1799,6 +1799,9 @@ PicoQuicTransport::SendStreamBytes(DataContext* data_ctx, std::uint64_t stream_i
     }
 
     bool should_reset = false;
+// tx_data is still locked here, but the analyzer can't see it, so skip the check.
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wthread-safety-analysis"
```
GhostofCookie: That sounds concerning, should we actually address it somehow? Why can't the analyzer see it?

src/transport_picoquic.cpp
```diff
@@ -1799,6 +1799,9 @@ PicoQuicTransport::SendStreamBytes(DataContext* data_ctx, std::uint64_t stream_i
...
     defer({
         const bool empty = stream_ctx.tx_data->Empty() && stream_ctx.tx_object == nullptr;
```
GhostofCookie, line 1803: Worth considering that the lock of the function scope does not apply in here, so we should probably lock again in here.

GhostofCookie: Or maybe it does...

GhostofCookie: Maybe this should actually go into the defer itself, if defer is obfuscating the locking from the analyzer?

RichLogan: Yeah, I don't the analyser can "see" outside the defer block, although I think the lock is held. Annoying it seems a pragma can't go in the defer itself either.

GhostofCookie: Hmm... yeah, the defer should be first on the stack, then the lock, so in theory the lock should be valid.

GhostofCookie: I guess we aren't helping the analyzer here. Maybe this pragma can go inside the destructor of DeferType

GhostofCookie (dismissed):

include/quicr/defer.h
```diff
@@ -7,7 +7,11 @@
 
 #define DEFER_CONCAT(a, b) DEFER_CONCAT_INNER(a, b)
 #define DEFER_CONCAT_INNER(a, b) a##b
-#define defer(n) quicr::DeferType DEFER_CONCAT(defer_, __LINE__)([&] { n; })
+#define defer(n)                                                                                                       \
+    quicr::DeferType DEFER_CONCAT(defer_, __LINE__)([&] {                                                              \
+        _Pragma("clang diagnostic push") _Pragma("clang diagnostic ignored \"-Wthread-safety-analysis\"") n;           \
+        _Pragma("clang diagnostic pop")                                                                                \
```
GhostofCookie: Fancy

GhostofCookie (dismissed):

GhostofCookie (approved):

PR #848, opened 2026-06-05
First pass of statically typed draft 18 messages

RichLogan:
Add type definitions for draft 18 control messages. 

More to be done - first-class parameters, uint8_ts to bools, couple more variant messages to unwrap etc but this is the main shape of all of them. 

CI failure is unrelated, needs to be fixed on draft-18 first. Codex errors are complaining that these messages don't match the draft-16 usage...obviously.


TimEvens (approved):
Much cleaner now.

PR #853, opened 2026-06-08
Draft 18 publish

RichLogan:


include/quicr/detail/control_messages/parameters.h
```diff
@@ -0,0 +1,98 @@
...
+    /// Must all be known, must not be duplicated (except auth).
+    inline void ValidateParameters(const Parameters& params, std::initializer_list<ParameterType> allowed)
+    {
+        std::vector<ParameterType> seen;
+        for (const auto& kvp : params) {
+            if (std::ranges::find(allowed, kvp.type) == allowed.end()) {
+                throw ProtocolViolationException("Parameter not valid for this message type");
+            }
+
+            if (kvp.type != ParameterType::kAuthorizationToken) {
```
GhostofCookie: Why?

src/transport.cpp
```diff
@@ -3544,53 +3545,43 @@ namespace quicr {
                 const auto parameters = messages::Message::ParseField<messages::Parameters>(msg_bytes);
                 const auto track_extensions = messages::Message::ParseField<messages::TrackExtensions>(msg_bytes);
 
-                auto tfn = FullTrackName{ track_namespace, track_name };
-                auto th = TrackHash(tfn);
-                conn_ctx.recv_req_id[request_id] = { .track_full_name = tfn,
+                // TODO: Fill all these in.
+                ValidateParameters(
+                  parameters, { ParameterType::kLargestObject, ParameterType::kForward, ParameterType::kGroupOrder });
+                const Publish publish{ .full_track_name = { track_namespace, track_name },
```
GhostofCookie: As a type name, `Publish` doesn't really evoke what it actually is. Before with the `Attribute` suffix at least it was clear by the name that it was data of some kind.

RichLogan: We can change it back to attributes, it's less of a diff too. I think I over-rotated on nuking the attributes when I should just rewrite their content.

include/quicr/detail/control_messages/parameters.h
RichLogan: Nothing is allowed to be duplicated except for auth tokens which are.

include/quicr/detail/parameters.h
```diff
@@ -0,0 +1,98 @@
...
+            value != static_cast<std::uint8_t>(GroupOrder::kDescending)) {
+            throw ProtocolViolationException("GROUP_ORDER parameter must be Ascending or Descending");
+        }
+        return static_cast<GroupOrder>(value);
+    }
+
+    inline std::optional<std::uint64_t> ResolveExpires(const Parameters& params)
+    {
+        const auto expires = params.GetOptional<std::uint64_t>(ParameterType::kExpires);
+        return (expires.has_value() && expires.value() != 0) ? expires : std::nullopt;
```
GhostofCookie, line 76: This is what GetOptional does.

include/quicr/detail/parameters.h
```diff
@@ -0,0 +1,98 @@
...
+    }
```
GhostofCookie, line 77:
```suggestion
    inline std::optional<std::uint64_t> ResolveExpires(const Parameters& params)
    {
        return params.GetOptional<std::uint64_t>(ParameterType::kExpires);
    }
```

include/quicr/detail/parameters.h
RichLogan: The reason I did it like this was so 0 expiry becomes nullopt rather than a value of 0 because 0 on the wire means it doesn't expire rather than it expires in 0 time.

>    If the EXPIRES parameter is 0 or is not present in a message, the
>    subscription does not expire or expires at an unknown time.

include/quicr/detail/control_messages/parameters.h
GhostofCookie: Why?

RichLogan: Can I interest you in a §? 😂

src/transport.cpp
```diff
@@ -607,9 +602,11 @@ namespace quicr {
...
          */
         auto params = Parameters{}
-                        .Add(ParameterType::kSubscriberPriority, priority)
-                        .AddOptional(ParameterType::kGroupOrder, group_order)
-                        .Add(ParameterType::kForward, forward);
+                        .AddOptional(ParameterType::kSubscriberPriority, priority)
+                        .AddOptional(ParameterType::kGroupOrder, group_order);
+        if (!forward) {
+            params.Add(ParameterType::kForward, forward);
+        }
```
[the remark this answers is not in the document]
RichLogan: We'll get to pubok next.

src/transport.cpp
```diff
@@ -3544,53 +3547,43 @@ namespace quicr {
                 const auto parameters = messages::Message::ParseField<messages::Parameters>(msg_bytes);
                 const auto track_extensions = messages::Message::ParseField<messages::TrackExtensions>(msg_bytes);
 
-                auto tfn = FullTrackName{ track_namespace, track_name };
-                auto th = TrackHash(tfn);
-                conn_ctx.recv_req_id[request_id] = { .track_full_name = tfn,
+                // TODO: Fill all these in.
+                ValidateParameters(
+                  parameters, { ParameterType::kLargestObject, ParameterType::kForward, ParameterType::kGroupOrder });
```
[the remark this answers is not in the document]
RichLogan: See the TODO that's right there.

include/quicr/detail/track_properties.h
```diff
@@ -0,0 +1,53 @@
...
+        }
+        return *value == 1;
+    }
+
+    inline std::uint8_t ResolveDefaultPublisherPriority(const TrackExtensions& properties)
+    {
+        return properties.GetOptional<std::uint8_t>(ExtensionType::kDefaultPublisherPriority).value_or(128);
+    }
+
+    inline std::optional<std::uint64_t> ResolveMaxCacheDuration(const TrackExtensions& properties)
```
RichLogan, line 43: idk if we really need these 0 logic ones or if it's worth it for the consistency of parameter parsing.

PR #860, opened 2026-06-10
Rename Transport to Session

GhostofCookie:


RichLogan (approved):
I haven't checked every line, but I agree with this change!

PR #861, opened 2026-06-11
Cleanup using stmts

GhostofCookie:


src/transport_picoquic.cpp
```diff
@@ -599,7 +599,7 @@ DefaultWebTransportCallback(picoquic_cnx_t* cnx,
         return -1;
     }
 
-    auto conn_id = reinterpret_cast<TransportConnId>(cnx);
+    auto conn_id = reinterpret_cast<std::uint64_t>(cnx);
```
RichLogan, line 602: Not this change, but reading this it occurs to me this is the pointer size and in theory it should be uintptr_t? But we don't want that in the API I don't think!

RichLogan (approved):
LGTM!

There are a couple of places where we might now want to alter a variable name to make it more obvious what it is (e.g it's a group), but I think we can just deal with them as we see them going forwards. 

For example `messages::RequestID rid` was sort of okay but I'd prefer `std::uint64_t request_id` over `std::uint64_t rid`.

GhostofCookie:
> LGTM!
> 
> There are a couple of places where we might now want to alter a variable name to make it more obvious what it is (e.g it's a group), but I think we can just deal with them as we see them going forwards.
> 
> For example `messages::RequestID rid` was sort of okay but I'd prefer `std::uint64_t request_id` over `std::uint64_t rid`.

Absolutely! I thought about it, but worried it would get a bit messy. I can make a follow up PR that addresses those.

PR #872, opened 2026-06-23
QUIC semantics for control 

RichLogan:
Draft 18 control message -> quic semantics 

Needs #871 

Also moves unsubscribe to handler status rather than session level callback.

### DIFF
--- a/include/quicr/session.h
+++ b/include/quicr/session.h
@@ -931,7 +931,7 @@ namespace quicr {
         void OnStreamClosed(const ConnectionHandle& connection_handle,
                             std::uint64_t stream_id,
                             std::shared_ptr<StreamRxContext> rx_ctx,
-                            std::optional<DataContextId> data_ctx_id,
+                            std::optional<uint64_t> data_ctx_id,
                             StreamClosedFlag flag) override;
 
       private:
--- a/include/quicr/session.h
+++ b/include/quicr/session.h
@@ -25,6 +25,7 @@
 #include <atomic>
 #include <chrono>
 #include <map>
+#include <set>
 #include <span>
 #include <string>
 #include <string_view>
@@ -926,12 +927,14 @@ namespace quicr {
                                    DataContextId data_ctx_id,
                                    const QuicDataContextMetrics& quic_data_context_metrics) override;
 
+      protected:
         void OnStreamClosed(const ConnectionHandle& connection_handle,
                             std::uint64_t stream_id,
                             std::shared_ptr<StreamRxContext> rx_ctx,
-                            std::optional<uint64_t> data_ctx_id,
+                            std::optional<DataContextId> data_ctx_id,
                             StreamClosedFlag flag) override;
 
+      private:
         // -------------------------------------------------------------------------------------------------
         // End of transport handler/callback functions
         // -------------------------------------------------------------------------------------------------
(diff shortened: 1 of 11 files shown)


src/session.cpp
```diff
@@ -1611,8 +1602,7 @@ namespace quicr {
             return;
         }
 
-        SendFetchCancel(conn_it->second, sub_id.value());
-
+        // TODO: Cancel using QUIC here when FETCH migrated.
```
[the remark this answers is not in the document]
RichLogan: The whole of FETCH needs to be migrated, so this is intentionally deferred for now.

include/quicr/session.h
```diff
@@ -926,12 +916,14 @@ namespace quicr {
...
                                    const QuicDataContextMetrics& quic_data_context_metrics) override;
 
+      protected:
         void OnStreamClosed(const ConnectionHandle& connection_handle,
                             std::uint64_t stream_id,
                             std::shared_ptr<StreamRxContext> rx_ctx,
                             std::optional<uint64_t> data_ctx_id,
                             StreamClosedFlag flag) override;
 
+      private:
```
GhostofCookie: Preferably, this function signature would just move to an already established `protected` block.

test/integration_test/test_server.cpp
```diff
@@ -122,8 +131,12 @@ TestServer::SubscribeReceived(ConnectionHandle connection_handle,
...
 
     // Create a publish track handler to send objects to this subscriber
-    auto pub_track_handler = std::make_shared<TestPublishTrackHandler>(
-      track_full_name, TrackMode::kStream, subscribe_attributes.priority, ttl);
+    auto pub_track_handler =
+      std::make_shared<TestPublishTrackHandler>(track_full_name,
+                                                TrackMode::kStream,
+                                                subscribe_attributes.priority,
+                                                ttl,
+                                                std::static_pointer_cast<TestServer>(shared_from_this()));
```
GhostofCookie, line 139: Odd, shouldn't have to cast this, do you?

RichLogan: Yeah I think we do. It's going from a `shared_ptr<Session>` -> `weak_ptr<TestServer>`.

src/session.cpp
```diff
@@ -1192,6 +1167,18 @@ namespace quicr {
                                       std::uint64_t stream_id,
                                       StreamClosedFlag flag)
     {
+        // Incoming PUBNS requests are not handler based.
+        if (std::erase(conn_ctx.recv_publish_namespaces, request_id) > 0) {
+            if (client_mode_) {
+                PublishNamespaceDoneReceived(request_id);
+            } else {
+                PublishNamespaceDoneReceived(connection_handle, request_id);
```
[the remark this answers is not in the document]
RichLogan: This is a relay function we should remove. Opened #873

suhasHere:
Curious Any reason for this refactor to introduce session API ?

RichLogan:
@suhasHere I'm not sure which bit you mean? Do you mean the change here for Unsubscribe to go to the handler rather than at the top session level? Most of the other events are scoped to the handler / request they route to rather than a flat top-level callback, so this brings `Unsubscribe` inline with those. With request streams and the quic close semantics, we can more cleanly keep details of a MoQ "request" scoped to the internal state of the track handler that represents it.

GhostofCookie (approved):

PR #874, opened 2026-06-29
Refactor names and types

GhostofCookie:


TimEvens (approved):
Looks good, but we will likely need to change the name of `connection_id` to something else because it is confusing with QUIC connection IDs.

PR #878, opened 2026-07-01
Draft 18 laps fixes

TimEvens:
* Updates picoquic to latest master
* Restores make transport so that laps peering can directly work with the transport. Peering doesn't use MOQT
* Restores UnsubscribeReceived() callback so that laps and other servers can clearly be notified of a publish that needs to be closed out due to a subscribe bidir stream close
* Adds integration test for UnsubscribeReceived() callback call via CloseRequestHandler()
* Resolves locking issues with OnStreamClose() calling CloseRequestHandler() with callbacks


src/session.cpp
```diff
@@ -1144,7 +1105,7 @@ namespace quicr {
                 [[fallthrough]];
             case SubscribeTrackHandler::Status::kOk:
                 try {
-                    if (send_unsubscribe && not handler.IsPublisherInitiated() && not conn_ctx.closed) {
+                    if (not handler.IsPublisherInitiated() && not conn_ctx.closed) {
```
RichLogan, line 1108: Looking at this, I think our stream closure here might both be incorrect here and also incorrect previously in terms of ensuring both sides of a bidir stream are correctly closed out? We can add a test for this and deal with it subsequently though.

test/integration_test/integration_test.cpp
```diff
@@ -400,6 +400,58 @@ TEST_CASE("Integration - Unsubscribe resets the subscribe request stream")
...
+        REQUIRE(sub_future.wait_for(kDefaultTimeout) == std::future_status::ready);
+        const auto request_id = sub_future.get().request_id;
+        REQUIRE(WaitFor([&handler]() { return handler->GetStatus() == SubscribeTrackHandler::Status::kOk; }));
+        REQUIRE(handler->GetRequestStreamId().has_value());
+        const auto request_stream_id = handler->GetRequestStreamId().value();
+
+        CHECK_NOTHROW(client->UnsubscribeTrack(handler));
+
+        REQUIRE(WaitFor([&]() { return server->WasStreamReset(request_stream_id).has_value(); }));
+        CHECK(server->WasStreamReset(request_stream_id) == true);
```
GhostofCookie:
```suggestion
        CHECK(server->WasStreamReset(request_stream_id));
```

test/integration_test/test_server.cpp
```diff
@@ -277,6 +277,20 @@ TestServer::StandaloneFetchReceived(const std::uint64_t connection_id,
...
+    std::lock_guard lock(state_mutex_);
+    if (unsubscribe_received_promise_.has_value()) {
+        const auto handler_type =
+          expected_unsubscribe_handler_type_.value_or(UnsubscribeReceivedDetails::HandlerType::kSubscribeTrack);
+        unsubscribe_received_promise_->set_value({ .request_id = request_id, .handler_type = handler_type });
+        unsubscribe_received_promise_.reset();
+        expected_unsubscribe_handler_type_.reset();
+    }
+    (void)connection_id;
+}
```
GhostofCookie:
```suggestion
TestServer::UnsubscribeReceived(const std::uint64_t, const uint64_t request_id)
{
    std::lock_guard lock(state_mutex_);
    if (unsubscribe_received_promise_.has_value()) {
        const auto handler_type =
          expected_unsubscribe_handler_type_.value_or(UnsubscribeReceivedDetails::HandlerType::kSubscribeTrack);
        unsubscribe_received_promise_->set_value({ .request_id = request_id, .handler_type = handler_type });
        unsubscribe_received_promise_.reset();
        expected_unsubscribe_handler_type_.reset();
    }
}
```

GhostofCookie (approved):

RichLogan (approved):

src/session.cpp
TimEvens: Let's revisit

PR #879, opened 2026-07-06
Move metrics callback to notify thread

RichLogan:
`OnConnectionMetricsSampled` was being called directly on the transport thread, rather than delegated to the notifier thread. This meant two things: firstly a client could in theory stall the transport, which would be bad, and secondly any session level state would be racing between the two threads (notify and transport). The time is already encoded in the structure, so a delayed metrics callback is no issue. I take a copy of what we need so we can continue to immediately reset period on the originating thread. 

Because the notify queue is lossy, we want to make sure a metrics callback will never pop a "real" callback. I also changed it to fire all the metrics events in one delegation (one queue entry) rather than N at a time, to reduce the chance of that happening.

### DIFF
--- a/src/transport_picoquic.cpp
+++ b/src/transport_picoquic.cpp
@@ -2200,19 +2200,28 @@ PicoQuicTransport::EmitMetrics()
 {
     for (auto& [conn_id, conn_ctx] : conn_context_) {
         const auto sample_time = std::chrono::system_clock::now();
+        const bool queue_space = cbNotifyQueue_.Size() < (tconfig_.callback_queue_size * 3) / 4;
 
-        cbNotifyQueue_.Push([this, sample_time, conn_id = conn_id, metrics = conn_ctx.metrics]() {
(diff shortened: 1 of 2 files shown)


src/transport_picoquic.cpp
```diff
@@ -2189,14 +2200,28 @@ PicoQuicTransport::EmitMetrics()
 {
     for (auto& [conn_id, conn_ctx] : conn_context_) {
         const auto sample_time = std::chrono::system_clock::now();
+        const bool queue_space = cbNotifyQueue_.Size() < (tconfig_.callback_queue_size * 3) / 4;
```
RichLogan, line 2203: Magic threshold here but codex raised a good point that we don't want to ever be in a situation where a real data callback is dropped in favour of a metrics one (that should be self-healing anyway).

GhostofCookie (approved):

TimEvens (approved):
LGTM

PR #880, opened 2026-07-06
Remove RequestOkReceived from public API

RichLogan:
Since we're already leveraging `friend`, we can just hide this implementation detail from the public API - the handler status and associated callback is what they should observe. 

This was already the case for namespace handlers.


TimEvens (approved):
LGTM

PR #885, opened 2026-07-08
Hide MoQ request update implementation detail

RichLogan:
Redesigns the public facing API to hide away MoQ request update internals, e.g: parameters. We can safely do this because libquicr must understand a parameter in order to parse it, and so we can work in "resolved" concepts like updated forward state or priority instead of kicking a parameter blob back to the app.  

The new flow looks like this:

- Internally, a REQUEST_UPDATE arrives with some parameters. 
- Internally, Session forwards this to the respective handler's internal request update callback.
- Internally, this callback parses the parameters and validates them according to the per-request rules. It then updates the internal state of the handler as appropriate (e.g the forward state). 
- Publicly, the handler notifies the app it has been updated (right now, through the existing Status callback). 
- At this point, the app can inspect the new state, and make a decision about whether to accept or reject the update. We can do this *after* setting the state because a rejection is terminal, so I figured we can avoid dealing with pending state since it's going to either be accepted or ended anyway. This may need some more thought, we could keep it pending until it's accepted.  
- The app then calls `ResolveRequestUpdate` to get libquicr to emit the right REQUEST_UPDATE_OK or REQUEST_ERROR. It can do this synchronously in the callback, or asynchronously later on (but it needs to understand the next bit).

The other thing to understand is that request updates have changed slightly, they're now only identifiable to each other based on the arrival ordering on the request stream, so they need to be resolved by the app in receipt order (FIFO). Multiple can be in-flight / pending at once as well, and the app resolve callbacks could be async, hence the atomic pending dance. It is up to the app to manage to resolve ordering though. `MAX_REQUEST_UPDATES` will come into play here later in draft-19. 

One thing that we might want to change is have updated be a dedicated callback instead of driven from status. We may also want to callback *what* properties have changed. 

You'll see in the default status callback I now resolve a request update blindly. This is a bit odd, but I do this because it allows the non-subclassed handlers to work out the box in the straight forward simple case, rather than falling over on any update. Anyone doing really work should subclass these anyway, so it seemed like the right thing to do.

### DIFF
--- a/include/quicr/session.h
+++ b/include/quicr/session.h
@@ -312,18 +312,6 @@ namespace quicr {
             std::optional<Bytes> error_reason;
         };
 
-        struct RequestUpdateResponse
-        {
-            struct Error
-            {
-                const messages::ErrorCode error_code;
-                const std::chrono::milliseconds retry_interval;
-                const std::string reason;
-            };
-            const std::optional<Error> error;
-            const messages::Parameters params;
-        };
-
         // --BEGIN RESOLVE METHODS ---------------------------------------------------------------------------
         /** @name Resolve Methods
          *      Methods to accept or reject inbound requests. Most are used in server mode; `ResolveSubscribe()`
@@ -431,17 +419,6 @@ namespace quicr {
         void ResolvePublishNamespaceDone(std::uint64_t connection_id,
                                          std::uint64_t request_id,
                                          const std::vector<std::uint64_t>& subscribers);
-
-        /**
-         * @brief Accept or reject a request update
-         *
-         * @param connection_id         Source connection ID
-         * @param request_id            Request being updated
-         * @param response              Request update response
-         */
-        void ResolveRequestUpdate(std::uint64_t connection_id,
-                                  uint64_t request_id,
-                                  const RequestUpdateResponse& response);
 
         /**
          * @brief Accept or reject track status that was received
@@ -1202,6 +1179,8 @@ namespace quicr {
 
         std::uint64_t ResponseDataContext(const ConnectionContext& conn_ctx, std::uint64_t request_id) const;
 
+        void ResolveRequestUpdate(const BaseTrackHandler& handler, const std::optional<RequestError>& error);
+
         /*===================================================================*/
         // Track Status
         /*===================================================================*/
@@ -1310,6 +1289,7 @@ namespace quicr {
         std::shared_ptr<timeq::tick_service> tick_service_;
         std::shared_ptr<ITransport> quic_transport_; // **MUST** be last for proper order of destruction
 
+        friend class BaseTrackHandler;
         friend class PublishTrackHandler;
         friend class PublishFetchHandler;
         friend class SubscribeTrackHandler;
--- a/include/quicr/session.h
+++ b/include/quicr/session.h
@@ -1179,7 +1179,12 @@ namespace quicr {
 
         std::uint64_t ResponseDataContext(const ConnectionContext& conn_ctx, std::uint64_t request_id) const;
 
-        void ResolveRequestUpdate(const BaseTrackHandler& handler, const std::optional<RequestError>& error);
+        std::shared_ptr<BaseTrackHandler> ResolveRequestUpdate(const BaseTrackHandler& handler,
+                                                               const std::optional<RequestError>& error);
+
+        void DispatchRequestUpdateCompletion(std::shared_ptr<BaseTrackHandler> handler,
+                                             bool rejected,
+                                             bool apply_next) const;
 
         /*===================================================================*/
         // Track Status
--- a/include/quicr/transport.h
+++ b/include/quicr/transport.h
@@ -12,6 +12,7 @@
 #include <any>
 #include <chrono>
 #include <cstdlib>
+#include <functional>
 #include <memory>
 #include <mutex>
 #include <optional>
@@ -438,6 +439,11 @@ namespace quicr {
         virtual void SetRemoteDataCtxId(std::uint64_t conn_id,
                                         std::uint64_t data_ctx_id,
                                         std::uint64_t remote_data_ctx_id) = 0;
+
+        /**
+         * @brief Run a callback on the notifier thread.
+         */
+        virtual void Dispatch(std::function<void()> callback) = 0;
 
         /**
          * Enqueue flags
--- a/src/base_track_handler.cpp
+++ b/src/base_track_handler.cpp
@@ -1,5 +1,6 @@
 #include "quicr/handlers/base_track_handler.h"
 #include "quicr/session.h"
+#include <spdlog/spdlog.h>
 
 namespace quicr {
     void BaseTrackHandler::SetTransport(std::shared_ptr<Session> transport)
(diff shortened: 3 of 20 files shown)


src/base_track_handler.cpp
```diff
@@ -12,6 +12,24 @@ namespace quicr {
         return transport_;
     }
 
+    void BaseTrackHandler::ResolveRequestUpdate(const std::optional<quicr::RequestError>& error)
+    {
+        // Consume a pending request.
+        auto pending = pending_request_updates_.load(std::memory_order_acquire);
+        while (pending != 0) {
+            if (pending_request_updates_.compare_exchange_weak(
+                  pending, pending - 1, std::memory_order_acq_rel, std::memory_order_acquire)) {
```
GhostofCookie: Not the most idiomatic thing to read but probably works fine.

GhostofCookie (approved):

RichLogan:
We discussed this, and will move to a stack of requests, and apply pending ones once the outstanding one has been resolved.

src/base_track_handler.cpp
```diff
@@ -12,6 +13,99 @@ namespace quicr {
...
+        }
+
+        // Get the session for this request.
+        const auto session = GetTransport().lock();
+        if (!session) {
+            ClearRequestUpdates();
+            return;
+        }
+
+        auto live_handler = session->ResolveRequestUpdate(*this, error);
```
GhostofCookie, line 59: This seems backwards... how are we calling into this handler method? Shouldn't Session have called us to get here? Seems odd to callback to session when session should have gotten us here.

src/session.cpp
```diff
@@ -2744,42 +2744,79 @@ namespace quicr {
...
 
-        SPDLOG_LOGGER_DEBUG(logger_, "Request Updated resolve req_id: {}", request_id);
+        return live_handler.handler;
+    }
+
+    void Session::DispatchRequestUpdateCompletion(std::shared_ptr<BaseTrackHandler> handler,
+                                                  bool rejected,
+                                                  bool apply_next) const
+    {
+        assert(!(rejected && apply_next));
```
GhostofCookie, line 2803: 👀

include/quicr/session.h
```diff
@@ -1197,6 +1174,13 @@ namespace quicr {
 
         std::uint64_t ResponseDataContext(const ConnectionContext& conn_ctx, std::uint64_t request_id) const;
 
+        std::shared_ptr<BaseTrackHandler> ResolveRequestUpdate(const BaseTrackHandler& handler,
+                                                               const std::optional<RequestError>& error);
+
+        void DispatchRequestUpdateCompletion(std::shared_ptr<BaseTrackHandler> handler,
+                                             bool rejected,
+                                             bool apply_next) const;
```
GhostofCookie, line 1182: Given that this method is not passed a callback to 'dispatch', prefer not naming it as such. What step does this function fulfill? Perhaps `CompleteRequestUpdate` or `RequestUpdatedCompleted`? Or maybe even `FinishRequestUpdate` or even simply `RequestUpdated`

include/quicr/transport.h
```diff
@@ -441,6 +442,11 @@ namespace quicr {
                                         std::uint64_t data_ctx_id,
                                         std::uint64_t remote_data_ctx_id) = 0;
 
+        /**
+         * @brief Run a callback on the notifier thread.
+         */
+        virtual void Dispatch(std::function<void()> callback) = 0;
```
GhostofCookie, line 448: While not technically a bad name, it feels disingenuous to what it does. Since it queues the method for the notify thread, wouldn't `Notify` be more apt? From the name alone I'd have no idea what or where this was going to dispatch. I have to come to the documentation to find out, rather than it being idiomatic from where it is called. Or maybe the more descriptive name would be `DispatchNotify` or something along those lines.

GhostofCookie: Actually, come to think, in PicoQuicTransport we have the RunPQFunction which queues on the runner thread. Maybe we can make some generic dispatching method where we pass an enum to specify which thread we want. Something like:
```cpp
enum class TransportThread
{
    Current,
    Notify,
    Runner,
};

void DispatchTo(TransportThread thread, std::function<void()>&& callback);
```

src/session.cpp
```diff
@@ -2744,42 +2744,79 @@ namespace quicr {
...
+            std::lock_guard lock(state_mutex_);
+            const auto conn_it = connections_.find(connection_id);
+            if (conn_it == connections_.end() || conn_it->second.closed) {
+                SPDLOG_LOGGER_WARN(logger_, "ResolveRequestUpdate: Connection doesn't exist");
+            } else {
+                const auto handler_it = conn_it->second.request_handlers.find(*request_id);
+                if (handler_it == conn_it->second.request_handlers.end()) {
+                    SPDLOG_LOGGER_WARN(logger_, "ResolveRequestUpdate: Stale handler");
+                } else if (handler_it->second.handler.get() != &handler) {
+                    SPDLOG_LOGGER_WARN(logger_, "ResolveRequestUpdate: Handler has been replaced");
```
GhostofCookie, line 2769: That's freaky, what does that mean if the handler was replaced? What happens to the handler that is a reference? Why are we not operating on the handler we have passed in?

src/session.cpp
```diff
@@ -3659,67 +3699,51 @@ namespace quicr {
...
+                        new_group_request = { sub_ctx_it->second.track_full_name, *new_group_request_id };
                     }
                 }
 
-                SendRequestUpdateOk(conn_ctx, data_ctx_id, std::nullopt, std::nullopt);
+                // Fire callbacks.
+                lock.unlock();
+                handler.handler->RequestUpdateReceived(parameters);
+                if (new_group_request.has_value()) {
+                    NewGroupRequested(new_group_request->first, new_group_request->second);
```
GhostofCookie, line 3745: Seems odd to have made this into a pair to then use the elements separately here.

src/subscribe_track_handler.cpp
```diff
@@ -12,6 +12,13 @@
 
 namespace quicr {
 
+    void SubscribeTrackHandler::StatusChanged(Status status)
+    {
+        if (status == Status::kSubscriptionUpdated) {
+            ResolveRequestUpdate();
+        }
+    }
```
GhostofCookie, line 20: What happens if a derived class overrides this method and doesn't know that they need the Super class' functionality?

test/integration_test/integration_test.cpp
```diff
@@ -143,6 +144,60 @@ class CallbackPublishTrackHandler final : public PublishTrackHandler
...
+                                    std::uint32_t default_ttl)
+      : TestPublishTrackHandler(full_track_name, TrackMode::kStream, default_priority, default_ttl)
+    {
+    }
+
+    void StatusChanged(Status status) override
+    {
+        if (status != Status::kSubscriptionUpdated) {
+            return;
+        }
```
GhostofCookie, line 161: About my earlier comment, here we see that in a derived class we have explicitly lost the functionality the parent class defined. ResolveRequestUpdate will not be called now.

GhostofCookie: I see it called lower down. Is it up to the library to define how it is resolved, or is it up to the app?

test/integration_test/integration_test.cpp
```diff
@@ -143,6 +144,60 @@ class CallbackPublishTrackHandler final : public PublishTrackHandler
...
+        {
+            std::lock_guard lock(mutex_);
+            --callback_depth_;
+        }
+    }
+
+    std::size_t UpdateCount() const
+    {
+        std::lock_guard lock(mutex_);
+        return update_count_;
```
GhostofCookie, line 185: Couldn't update_count_ just be an atomic?

test/integration_test/integration_test.cpp
RichLogan: I tried to touch on this in the description. Basically the app must do it, but I made the base auto-resolve so that you could still use the base handlers in a pinch. 

I am leaning towards changing this drastically though, because StatusChanged(kUpdated) is not enough info really. For example, new group request has the group ID, that is lost today.

PR #887, opened 2026-07-09
Subgroup original first object support

RichLogan:
Adds support for draft-18's first object bit (when true, the first object on that subgroup is the original first object on the subgroup, there are no priors in the past).

### DIFF
--- a/include/quicr/messages/messages.h
+++ b/include/quicr/messages/messages.h
@@ -111,17 +111,22 @@ namespace quicr::messages {
         // If true, the priority field is omitted and the subgroup inherits the publisher priority.
         // If false, the priority field is present and the subgroup has its own priority.
         const bool default_priority;
+        // If true, this is the original first object of this subgroup.
+        // If false, this may be the first object on the subgroup, but there are missing priors.
+        const bool first_object;
 
         static constexpr std::uint8_t kExtensionsBit = 0x01;
         static constexpr std::uint8_t kSubgroupIdBit = 0x06;
         static constexpr std::uint8_t kEndOfGroupBit = 0x08;
         static constexpr std::uint8_t kDefaultPriorityBit = 0x20;
+        static constexpr std::uint8_t kFirstObjectBit = 0x40;
 
         explicit constexpr StreamHeaderProperties(const std::uint64_t type)
           : extensions(type & kExtensionsBit)
           , subgroup_id_mode(static_cast<SubgroupIdType>((type & kSubgroupIdBit) >> 1))
           , end_of_group(type & kEndOfGroupBit)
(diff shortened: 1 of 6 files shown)


src/publish_track_handler.cpp
```diff
@@ -32,7 +32,7 @@ namespace quicr {
                 if (stream_mode.has_value()) {
                     stream_mode_.emplace(*stream_mode);
                 } else {
-                    stream_mode_.emplace(true, messages::SubgroupIdType::kExplicit, false, false);
+                    stream_mode_.emplace(true, messages::SubgroupIdType::kExplicit, false, false, false);
```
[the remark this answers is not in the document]
RichLogan: I'm not sure how to feel about this. We cannot know, so maybe having this in stream properties and setting a default here (which is really just the default pattern) is the wrong move entirely. Originally I was thinking that a truthy default is actually the less likely case, but in terms of out of the box usage of libquicr maybe true *would* be a better default. Yeah I think I'm convinced.

include/quicr/messages/messages.h
```diff
@@ -155,12 +162,15 @@ namespace quicr::messages {
...
+            if (first_object) {
+                type |= kFirstObjectBit;
+            }
             return type;
         }
 
         static constexpr bool IsValid(const std::uint64_t type) noexcept
         {
-            if ((type & 0b11010000) != 0b00010000) {
+            if ((type & 0b10010000) != 0b00010000) {
```
GhostofCookie: Magic

include/quicr/messages/messages.h
```diff
@@ -111,17 +111,22 @@ namespace quicr::messages {
...
         const bool default_priority;
+        // If true, this is the original first object of this subgroup.
+        // If false, this may be the first object on the subgroup, but there are missing priors.
+        const bool first_object;
 
         static constexpr std::uint8_t kExtensionsBit = 0x01;
         static constexpr std::uint8_t kSubgroupIdBit = 0x06;
         static constexpr std::uint8_t kEndOfGroupBit = 0x08;
         static constexpr std::uint8_t kDefaultPriorityBit = 0x20;
+        static constexpr std::uint8_t kFirstObjectBit = 0x40;
```
GhostofCookie, line 123: Should these all be an enum?

GhostofCookie (approved):

include/quicr/messages/messages.h
RichLogan: I broke this out a bit so it's less magical

include/quicr/messages/messages.h
RichLogan: They're more individual but related constants than something that would ever be enumerated or casted to/from a type.

PR #888, opened 2026-07-09
Enable debug logging and fix bind publish track to register request_id by data_ctx_id

TimEvens:
Enables debug logging and fixes debug messages based on draft-18 changes


RichLogan (approved):

RichLogan:
For posterity those codex reviews are both unrelated and known (I think just because the PR was opened against base not draft-18 and they ran too early)

Issue #873, opened 2026-06-25
Remove relay functionality from PUBLISH_NAMESPACE_DONE


RichLogan:
It's a relay's responsibility, not libquicr's.

PR #892, opened 2026-07-10
Refactor session

GhostofCookie:


include/quicr/session.h
```diff
@@ -101,132 +102,154 @@ namespace quicr {
...
 
-        ~Session();
-
-        const std::shared_ptr<timeq::tick_service>& GetTickService() const noexcept { return tick_service_; }
+        virtual ~Session();
 
-        // -------------------------------------------------------------------------------------------------
-        // Lifecycle
-        // -------------------------------------------------------------------------------------------------
+        const std::shared_ptr<Connection>& GetConnection() const noexcept { return current_connection_; }
```
RichLogan, line 198: Is there a reason why we're returning this (and the others) as a const-ref to the shared_ptr and not just the shared_ptr itself?

src/session.cpp
```diff
@@ -2710,72 +2284,60 @@ namespace quicr {
...
+    void Session::ResolvePublishNamespaceDone(std::uint64_t request_id, const std::vector<std::uint64_t>& subscribers)
     {
-        for (const auto& sub_conn_handle : subscribers) {
-            auto it = connections_.find(sub_conn_handle);
-            if (it == connections_.end()) {
-                continue;
-            }
+        // TODO: Figure this out now that Session/Connections(s) are separated from each other. Maybe
+        // SessionManager?
+        SPDLOG_LOGGER_ERROR(logger_, "ResolvePublishNamespaceDone is unimplemented until we can coordinate Sessions");
```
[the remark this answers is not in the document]
RichLogan: I think for this one see #873. IMO removing it is actually the right thing - this functionality should be specifically implemented by a relay.

GhostofCookie:
@codex review

GhostofCookie:
@codex review

src/picoquic_connection.h
```diff
@@ -0,0 +1,202 @@
...
+        bool wt_h3_ctx_owned{ false };
+
+        /// Client mode: connection-specific authority (server:port)
+        std::string wt_authority;
+
+        /// WebTransport capsule accumulator for control stream message parsing
+        /// Used to parse CLOSE_WEBTRANSPORT_SESSION and other capsules
+        picowt_capsule_t wt_capsule{};
+
+        char peer_addr_text[45]{ 0 };
```
fluffy, line 227: Perhaps add some clarity on what this is

include/quicr/config.h
```diff
@@ -3,12 +3,38 @@
...
+     */
+    struct TransportConfig
+    {
+        std::string tls_cert_filename;               ///< QUIC TLS certificate to use
+        std::string tls_key_filename;                ///< QUIC TLS private key to use
+        uint32_t time_queue_init_queue_size{ 1000 }; ///< Initial queue size to reserve upfront
+        uint32_t time_queue_max_duration{ 2000 };    ///< Max duration for the time queue in milliseconds
+        uint32_t time_queue_bucket_interval{ 500 };  ///< The bucket interval in milliseconds
+        uint32_t time_queue_rx_size{ 1000 };         ///< Receive queue size
+        bool debug{ false };                         ///< Enable debug logging/processing
```
fluffy, line 23: would it make sense to have this do the usual INFO, WARN, DEBUG, ERROR levels ?

include/quicr/config.h
```diff
@@ -3,12 +3,38 @@
...
+        uint64_t quic_cwin_minimum{ 131072 };        ///< QUIC congestion control minimum size (default is 128k)
+        uint32_t quic_wifi_shadow_rtt_us{ 20000 };   ///< QUIC wifi shadow RTT in microseconds
+        uint64_t idle_timeout_ms{ 30000 };           ///< Idle timeout for transport connection(s) in milliseconds
+        bool use_reset_wait_strategy{ false };       ///< Use Reset and wait strategy for congestion control
+        bool use_bbr{ true };                        ///< Use BBR if true, NewReno if false
```
fluffy, line 27: could we make this an ENUM that sets the congestion strategy. This will make it easier to add new alg liek C4

fluffy (commented):
Few trivial comments put in - feel free to ignore. I'm not up to speed on this.

src/picoquic_connection.h
GhostofCookie: We're going to be doing a full documentation run after this goes through.

include/quicr/config.h
GhostofCookie: Possibly, that could fall under the #904 since we now hide spdlog variables. I believe this is a legacy variable that we set via command line args.

include/quicr/config.h
GhostofCookie: We certainly could. Would probably be good to do under a separate PR for clarity of purpose.

.github/workflows/cmake.yml
```diff
@@ -39,7 +39,7 @@ jobs:
     - name: Install Packages (macOS)
       if: matrix.os == 'macos-latest'
       run: |
-        brew install llvm go google-benchmark
+        HOMEBREW_NO_REQUIRE_TAP_TRUST=1 brew install llvm go google-benchmark
```
RichLogan: Do we still need `go`? Was this only for boringSSL?

src/connection.cpp
```diff
@@ -0,0 +1,105 @@
...
+    {
+        {
+            std::lock_guard lock(delegate_mutex_);
+            delegate_ = delegate;
+        }
+
+        if (delegate && status_ != Status::kConnecting) {
+            // Replay the current status in case transport notifications were delivered
+            // before the application delegate was attached.
+            OnStatusChanged(status_);
```
RichLogan, line 63: Pretty minor but in this case this callback is not fired on the notifier thread like the rest of the status notifications, which might make this potentially race-y.

.github/workflows/cmake.yml
GhostofCookie: Good point. Maybe we can get rid of it and speed this up

src/connection.cpp
GhostofCookie: Does it have to be on the notify thread? We can move it to there.

RichLogan (approved):

PR #900, opened 2026-07-21
Remove SafeStreamBuffer & mutex

RichLogan:
In looking at the thread safety of `SafeStreamBuffer` and realising it needed some work, I then realised we don't even use `SafeStreamBuffer`, or have cross-thread `StreamBufffer` usage at all. So it seemed rather than fix it, just remove it. 

If we do need a thread-safe stream buffer in the future, I would suggest using a lockable buffer like the pattern we have with the queues, in order to ensure safe access to spans during their lifetime.


GhostofCookie (approved):
Nice! Good to know we are safely using the buffer.

PR #901, opened 2026-07-22
Add SWIG to generate language bindings

GhostofCookie:


suhasHere:
Even though swig is a nice interface, but most of the modern design typically use 
pybind11 for python, cgo for Go and so on.

Most common issues with Swig has been verbose generated code and messy build systems

PR #903, opened 2026-07-22
Use stream buffer for control stream bytes

RichLogan:
Use `StreamBuffer` for incoming control and request stream bytes. This allows us to align processing of stream buffers across control, request and data-plane streams (the latter yet to be done) going forward. Incrementing a span makes the parsing easier to follow as well. 

Possibly there is a theoretical perf improvement here as well?

### DIFF
--- a/src/session.cpp
+++ b/src/session.cpp
@@ -1198,7 +1198,7 @@ namespace quicr {
         // Incoming PUBNS requests are not handler based.
         if (std::erase(conn_ctx.recv_publish_namespaces, request_id) > 0) {
             conn_ctx.recv_req_id.erase(request_id);
-            conn_ctx.ctrl_msg_buffer.erase(stream_id);
+            conn_ctx.stream_buffers.erase(stream_id);
 
             lock.unlock();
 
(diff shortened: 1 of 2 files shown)


src/session.cpp
```diff
@@ -1879,47 +1879,50 @@ namespace quicr {
...
+                                    stream_buffer.Size());
 
-                while (ctrl_msg_buffer.data.size() > 0) {
-                    if (ctrl_msg_buffer.data.size() < UintVar::Size(ctrl_msg_buffer.data.front())) {
+                while (!stream_buffer.Empty()) {
+                    auto message_view = stream_buffer.Data();
+
+                    // Type.
+                    const auto size = UintVar::Size(message_view.front());
+                    if (message_view.size() < size) {
```
GhostofCookie: Interesting, why not use StreamBuffer::Front?

RichLogan: No specific reason other than I guess to have all the parsing done on that one span in the same way. But true for this first front and size we could just use `stream_buffer` itself, and defer the span creation if make it past there.

GhostofCookie (approved):

PR #904, opened 2026-07-30
Remove Spdlog and add abstract logger interface

GhostofCookie:
Spdlog really shouldn't be an exposed dependency that other libraries magically have included for them. If we ever decided to change loggers randomly suddenly people wouldn't have it for themselves. Instead this just wraps it in our own logger interface that uses spdlog under the hood.

Also in this PR, some of the Trace logs were broken (we clearly never compile with Trace enabled enough), so those needed some fixing as well.

### DIFF
--- a/include/quicr/log.h
+++ b/include/quicr/log.h
@@ -5,7 +5,7 @@
 #include <format>
 namespace std_or_fmt = std;
 #else
-#include <fmt/core.h>
+#include <fmt/format.h>
 namespace std_or_fmt = fmt;
 #endif
 #include <memory>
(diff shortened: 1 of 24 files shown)


GhostofCookie:
@codex review

include/quicr/log.h
```diff
@@ -0,0 +1,107 @@
...
+        template<typename... Args>
+        void Log(std::source_location location,
+                 Level level,
+                 std::conditional_t<sizeof...(Args) == 0, std::string_view, std_or_fmt::format_string<Args...>> msg,
+                 Args&&... args)
+        try {
+            if constexpr (sizeof...(Args) == 0) {
+                Log(level, msg, location);
+            } else {
+                Log(level, std_or_fmt::vformat(msg.get(), std_or_fmt::make_format_args(args...)), location);
```
RichLogan, line 55: I wonder if we should have a current level API so we can check if we would actually log this prior to formatting?

RichLogan (approved):

PR #908, opened 2026-08-04
Stream buffer fix

RichLogan:
Accumulate fragmented stream start up until routable to handlers. For subscribe, this is track alias, for fetch, request ID. 

Pending non-routable bytes accumulate into the stream buffer into session, and are then move'd into the handler to parse. On non-stream start, the existing shared_ptr data path is taken to accumulate into the buffer now owned by the handler. On steam start, the initial event contains both the buffer with the routable bytes, and the original data ptrs, should a consumer (laps) want them for forwarding. libquicr itself doesn't use them, just provides them.


TimEvens (approved):
LGTM

PR #910, opened 2026-08-10
Implement draft-18 FETCH serialization

RichLogan:
- Make group order non-optional, defaulted to ascending for FETCH.

### DIFF
--- a/src/fetch_track_handler.cpp
+++ b/src/fetch_track_handler.cpp
@@ -10,6 +10,7 @@ namespace quicr {
     {
         if (not stream.buffer.AnyHasValue()) {
             stream.buffer.InitAny<messages::FetchHeader>();
+            serialization_state_ = messages::FetchObjectSerializationState(*GetGroupOrder());
         }
 
         auto& f_hdr = stream.buffer.GetAny<messages::FetchHeader>();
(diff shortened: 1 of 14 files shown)


src/fetch_track_handler.cpp
```diff
@@ -10,6 +10,7 @@ namespace quicr {
     {
         if (not stream.buffer.AnyHasValue()) {
             stream.buffer.InitAny<messages::FetchHeader>();
+            serialization_state_ = messages::FetchObjectSerializationState(*GetGroupOrder());
```
TimEvens, line 13: Why is GetGroupOrder() an optional if it's expected?

src/messages.cpp
```diff
@@ -324,22 +324,332 @@ namespace quicr::messages {
...
+        if (first_object) {
+            // Absolute value.
+            group_id_delta = headers.group_id;
+        } else if (group_id_delta_present) {
+            // Delta encode based on group order.
+            if (group_order_ == GroupOrder::kAscending) {
+                if (headers.group_id <= *prior_group_id_) {
+                    throw std::invalid_argument("FETCH groups are not in ascending order");
+                }
+                group_id_delta = headers.group_id - *prior_group_id_ - 1;
```
TimEvens, line 420: Can both headers.group_id and prior_group_id be zero?

src/messages.cpp
```diff
@@ -324,22 +324,332 @@ namespace quicr::messages {
...
+        if (!datagram && subgroup_mode == FetchSerializationProperties::SubgroupIdMode::kExplicit) {
+            object.subgroup_id = headers.subgroup_id;
+        }
+        if (object_id_delta_present) {
+            object.object_id_delta = group_id_delta_present ? headers.object_id : headers.object_id - *prior_object_id_;
+        }
+        if (priority_present) {
+            object.publisher_priority = priority;
+        }
+        object.extensions = headers.extensions;
```
TimEvens, line 468: wonder if we can avoid copies

src/messages.cpp
```diff
@@ -324,22 +324,332 @@ namespace quicr::messages {
...
+            const std::uint64_t object_id_delta = *object.object_id_delta;
+            if (first_object || object.group_id_delta.has_value()) {
+                // First or first of group has absolute value.
+                object_id = object_id_delta;
+            } else {
+                // Increment by given delta.
+                if (object_id_delta > std::numeric_limits<std::uint64_t>::max() - *prior_object_id_) {
+                    throw ProtocolViolationException("FETCH Object ID overflow");
+                }
+                object_id = *prior_object_id_ + object_id_delta;
```
TimEvens, line 548: we don't need +1?

src/fetch_track_handler.cpp
RichLogan: This is inheritance slightly getting in the way, it's still optional on the parent subscribe track handler which is what's being called here, even though in the fetch case it will never be nullopt.

src/messages.cpp
RichLogan: No, that check just above 420 would protect against that, as that would be true if they were both 0 and it would throw.

src/messages.cpp
RichLogan: No I don't think so, here's the relevant block, with this case bolded.

> When the Group ID Delta field is present, the Object ID is the value of Object ID Delta if present. **When the Group ID Delta field is not present, the Object ID is the prior Object's ID plus the Object ID Delta if present**. If Object ID Delta is not present, the Object ID is the prior Object's ID plus one, regardless of which group it belongs to. If the computed Object ID would be greater than 2^64-1, the Subscriber MUST close the Session with error 'PROTOCOL_VIOLATION'.

src/messages.cpp
RichLogan: I think that we could, but we probably want to do that in another pass, because we could actually make the same saving in the publish path too.

TimEvens (approved):
LGTM

PR #913, opened 2026-08-17
Stream closure cleanup

RichLogan:
I reviewed stream closures in a decent amount of detail, and had a few findings / fixes.

- Bidirs didn't close the peer direction properly, now they do. Added a new `kCancel` close operation to communicate the "RESET + STOP_SENDING" on each half action.
- Check close operation makes sense on each stream (you can't FIN/RESET a recv, you can't STOP_SENDING a send). 
- First class STOP_SENDING.
- Subgroup close doesn't accidentally close the request.
- Cleanup the... cleanup.
- REQUEST_ERROR can immediately FIN
- WT changes that showed up in tests, they handle some cleanup internally. 

More to do:

- TRACK_STATUS
- Error codes set and bubbled up
- The per request error handling on update failure.


src/session.cpp
```diff
@@ -341,18 +341,24 @@ namespace quicr {
...
-                SendTrackStatusOk(conn_it->second,
-                                  ResponseDataContext(conn_it->second, request_id),
-                                  subscribe_response.largest_location,
-                                  TrackExtensions());
+                SendTrackStatusOk(
+                  conn_it->second, *data_ctx_id, subscribe_response.largest_location, TrackExtensions());
                 break;
             }
             case RequestResponse::ReasonCode::kDoesNotExist:
                 SendRequestError(conn_it->second,
```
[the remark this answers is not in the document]
RichLogan: See #890

include/quicr/transport.h
```diff
@@ -169,12 +169,60 @@ namespace quicr {
...
+    // Events received on a stream.
     enum class StreamClosedFlag : uint8_t
     {
         kFin,
         kReset,
+        kStopSending,
     };
 
+    // Action to take on a stream.
+    enum class StreamOperation : uint8_t
```
GhostofCookie, line 154: What is fundamentally different between this and StreamClosed flag? Could they not be the same enum?

include/quicr/transport.h
```diff
@@ -169,12 +169,60 @@ namespace quicr {
...
+        kCancel,
+    };
+
+    /**
+     * Validate a close operation is correct for that stream.
+     * @param stream_id The stream we're going to close.
+     * @param is_server True if we're the server.
+     * @param close The type of close operation.
+     */
+    constexpr void CheckCloseStream(std::uint64_t stream_id, bool is_server, StreamOperation close)
```
GhostofCookie: Should this return a bool or an enum value?

include/quicr/transport.h
```diff
@@ -169,12 +169,60 @@ namespace quicr {
...
+                break;
+            case StreamOperation::kStopSending:
+                if (can_receive) {
+                    return;
+                }
+                break;
+            case StreamOperation::kCancel:
+                if (is_bidir) {
+                    return;
+                }
```
GhostofCookie, line 186: Are these if blocks just stubs for future features?

include/quicr/transport.h
RichLogan: I considered this, but one is what you can do to a steam, and one is what you can receive to a stream.

include/quicr/transport.h
RichLogan: It's more like an assertion, as it's pure programmer error, it just throws if it's wrong.

include/quicr/transport.h
RichLogan: No, it's the happy case where it's allowed, and anything else hits the exception. But I can invert it.

src/session.cpp
```diff
@@ -518,12 +522,12 @@ namespace quicr {
...
-                    ControlMessageType::kRequestError,
-                    error,
-                    UintVar(retry_interval.count()),
-                    AsOwnedBytes(reason));
+        // Send REQUEST_ERROR and potentially close the request stream.
+        messages::Message msg = messages::Message{}.PrependType(ControlMessageType::kRequestError).ReserveLength();
+        msg.Append(error);
+        msg.Append(UintVar(retry_interval.count()));
+        msg.Append(AsOwnedBytes(reason));
+        SendCtrlMsg(conn_ctx, data_ctx_id, msg.ToBytes(), close_stream);
```
GhostofCookie: Could the variadic SendCtrlMsg not be modified to take close_stream before the variadic args? That would avoid you having to build the messages yourself.

RichLogan: Yeah it could, iirc it was just that this is the only case where you want to set it to true, and you'd have to alter every call site if we did that, and we can't default it because of the variadic.

src/transport_picoquic.cpp
```diff
@@ -59,6 +59,15 @@ using namespace quicr;
 
 constexpr const char* kMoqtAlpn = "moqt-18";
 
+static bool
+IsStreamFullyClosed(const PicoQuicStream& stream)
```
GhostofCookie: This could probably be a first class method on the stream object itself.

src/transport_picoquic.cpp
```diff
@@ -2117,7 +2137,9 @@ PicoQuicTransport::RemoveClosedStreams(std::size_t shard_idx)
                 continue;
             }
 
-            if (stream->tx_data != nullptr) {
+            if (IsStreamFullyClosed(*stream)) {
+                connection->RemoveStream(stream_id);
```
GhostofCookie: Does this make the else branch unreachable?

src/transport_picoquic.cpp
```diff
@@ -2784,32 +2806,37 @@ PicoQuicTransport::CloseStream(const std::shared_ptr<Connection>& connection,
     }
 
     const auto pq_stream = std::static_pointer_cast<PicoQuicStream>(stream);
+    CheckCloseStream(pq_stream->GetStreamId(), is_server_mode, operation);
```
GhostofCookie, line 2797: If CloseStream does this check itself, is this redundant?

test/integration_test/integration_test.cpp
```diff
@@ -772,7 +786,9 @@ TEST_CASE("Integration - Unsubscribe resets the subscribe request stream")
     auto server = MakeTestServer(session_mgr);
 
     auto test_unsubscribe = [&](const std::string& protocol_scheme) {
-        auto [session, callbacks] = MakeTestClient(session_mgr, true, std::nullopt, protocol_scheme);
+        const auto client = MakeTestClient(session_mgr, true, std::nullopt, protocol_scheme);
+        const auto& session = client.first;
+        const auto& callbacks = client.second;
```
GhostofCookie:
```suggestion
        const auto& [session, callbacks] = MakeTestClient(session_mgr, true, std::nullopt, protocol_scheme);
```

test/integration_test/integration_test.cpp
```diff
@@ -873,6 +893,132 @@ TEST_CASE("Integration - CloseRequestHandler UnsubscribeReceived when client Uns
...
+{
+    auto test_rejection = [](const std::string& protocol_scheme) {
+        quicr::SessionManager session_mgr;
+        auto server = MakeTestServer(session_mgr);
+
+        // Setup to blanket reject the request.
+        server->SetSubscribeError(RequestErrorCode::kDoesNotExist, "Track does not exist");
+        const auto client = MakeTestClient(session_mgr, true, std::nullopt, protocol_scheme);
+        const auto& session = client.first;
+        const auto& callbacks = client.second;
```
GhostofCookie, line 906:
```suggestion
        const auto& [session, callbacks] = MakeTestClient(session_mgr, true, std::nullopt, protocol_scheme);
```

GhostofCookie (dismissed):

GhostofCookie (dismissed):

GhostofCookie (dismissed):

TimEvens (dismissed):
LGTM

GhostofCookie (approved):

PR #915, opened 2026-08-18
Improve data ctx

GhostofCookie:


GhostofCookie:
@codex Review

src/session.cpp
```diff
@@ -1244,7 +1237,7 @@ namespace quicr {
                                       th.track_fullname_hash);
                 }
 
-                pub_n_it->second->publish_data_ctx_id_ = 0;
+                pub_n_it->second->ResetPublishDataContext();
```
RichLogan, line 1240: We reset this here, but then re-lookup on 1252, which will always be nil in that case? Do we need to capture it here, reset the handler value, and pass the captured unbound context to delete so it can do cleanup? Otherwise I think it might be left dangling?

GhostofCookie:
Closing in preference of https://github.com/Quicr/libquicr/pull/923

PR #921, opened 2026-08-26
Use queue to mark active instead of function pointer

TimEvens:


RichLogan (dismissed):

GhostofCookie (approved):
Approved as verified working on LAPS.

PR #923, opened 2026-08-27
Replace DataContext with per-stream contexts.

GhostofCookie:
Streams become first-class shared objects owned by the connection, carrying both directions, and are what picoquic holds as its app stream context. Track handlers hold stream handles directly, so the flow-level DataContext indirection between a handler and the bytes it sends is gone, along with the dead reset-wait machinery.


GhostofCookie:
@codex Review

docs/stream-context-refactor.md
```diff
@@ -0,0 +1,314 @@
+# Stream Context Usage: Analysis and Path Forward
```
TimEvens: Do we need this doc?

TimEvens (approved):
LGTM

docs/stream-context-refactor.md
GhostofCookie: Nope!

PR #924, opened 2026-09-01
Thread safe connection delegate

RichLogan:
I think this needs a mutex because the weak_ptr instance *itself* can be unset by an app thread calling `Session::Disconnect` while notifiers are running.

Random aside, seems my mac's clang doesn't seem to have std::atomic<std::weak_ptr>> even though it's in C++20? Shame.


GhostofCookie (approved):
Looks good! Shame about atomic weak ptr not being supported. AppleClang is truly awful for supporting features.

PR #927, opened 2026-09-03
Add missing exception header

RichLogan:
main doesn't compile for me on `-DLINT=ON` without these

### DIFF
--- a/src/fetch_track_handler.cpp
+++ b/src/fetch_track_handler.cpp
@@ -1,6 +1,8 @@
 // SPDX-FileCopyrightText: Copyright (c) 2025 Cisco Systems
 // SPDX-License-Identifier: BSD-2-Clause
 
+#include <exception>
+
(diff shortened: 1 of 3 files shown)


src/fetch_track_handler.cpp
```diff
@@ -1,6 +1,8 @@
 // SPDX-FileCopyrightText: Copyright (c) 2025 Cisco Systems
 // SPDX-License-Identifier: BSD-2-Clause
 
+#include <exception>
```
GhostofCookie, line 4: System headers go last in the ordering of header blocks.

GhostofCookie (changes requested):

src/fetch_track_handler.cpp
RichLogan: boo I was expecting make format to move these for me

GhostofCookie (approved):

PR #928, opened 2026-09-03
Improve Callbacks

GhostofCookie:
Makes the request update callback match the Reply-Resolve pattern that session callbacks uses, and also moves stream.h back in src to avoid public use.


TimEvens (approved):
LGTM

PR #931, opened 2026-09-08
Upgrade sframe

RichLogan:
Instead of #930

### DIFF
--- a/examples/qclient/client.cpp
+++ b/examples/qclient/client.cpp
@@ -249,7 +249,16 @@ class MySubscribeTrackHandler : public quicr::SubscribeTrackHandler
     {
         quicr::Bytes pt(data.size());
         if (qclient_vars::mls_ctx.has_value()) {
-            data = qclient_vars::mls_ctx->unprotect(pt, data, {});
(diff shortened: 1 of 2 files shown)


examples/qclient/client.cpp
```diff
@@ -249,7 +249,16 @@ class MySubscribeTrackHandler : public quicr::SubscribeTrackHandler
     {
         quicr::Bytes pt(data.size());
         if (qclient_vars::mls_ctx.has_value()) {
-            data = qclient_vars::mls_ctx->unprotect(pt, data, {});
+            auto plaintext = qclient_vars::mls_ctx->unprotect(pt, data, {});
+            if (plaintext.is_err()) {
+                const auto error = plaintext.error();
+                QUICR_LOGGER_WARN(qclient_vars::logger,
```
GhostofCookie, line 255: Shouldn't this be an error?

GhostofCookie (approved):
Looks good, though one of the messages is a warning instead of error?

examples/qclient/client.cpp
RichLogan: I guess if we're assuming a symmetrical working example yes. But if you received something you couldn't decrypt it might not be your fault?

PR #933, opened 2026-09-09
Add windows compatibility.

GhostofCookie:


include/quicr/containers/priority_queue.h
```diff
@@ -248,7 +248,7 @@ namespace quicr {
             /*
              * Initialize free time queues with two starting entries.
              */
-            for (int i = free_tqueues_.size(); i < kMinFreeTimeQueues; ++i) {
+            for (int i = static_cast<int>(free_tqueues_.size()); i < kMinFreeTimeQueues; ++i) {
```
RichLogan, line 251: Extremely minor but I guess `k[Min/Max]FreeTimeQueues` should be `std::size_t` and then all of this would fall out more nicely. They're only used relative to sizes anyway.

include/quicr/utilities/defer.h
```diff
@@ -16,11 +16,19 @@
 #define DEFER_POP
 #endif
 
+#ifdef _WIN32
+#define defer(n)                                                                                                       \
+    quicr::ScopeGuard DEFER_CONCAT(defer_, __LINE__)([&]() {                                                           \
```
RichLogan: If we used FORCE_INLINE like elsewhere then would we get away with a single definition here?

RichLogan (approved):
LGTM

RichLogan (dismissed):
LGTM

include/quicr/utilities/defer.h
GhostofCookie: Windows did not like attributes in the lambda so it didn't really work. It's more of a hint to the compiler anyway, the defer should be inlining it by default with our usage anyway.

RichLogan (approved):

PR #938, opened 2026-09-15
Move shutdown status

RichLogan:
Move this slightly further down ensures that when the status callback fires, the tracks have already been closed. This  also makes non locally initiated shutdowns work in quicr-mac like they used to.


TimEvens (approved):
LGTM

PR #939, opened 2026-09-15
Remove shared bytes for received data

GhostofCookie:
Replaces the queue of shared ptr of Bytes with a StreamBuffer directly, avoiding the need for subscribe handlers to have one, and removes all knowledge of a stream from track handlers.


GhostofCookie:
@codex Review

GhostofCookie:
@codex Review

include/quicr/handlers/subscribe_track_handler.h
```diff
@@ -418,25 +402,18 @@ namespace quicr {
...
-         * @details The stream buffer may contain zero or more objects and/or incomplete data.
-         * @param stream Context of stream with newly arrived data.
+         * Delta state for decoding the objects of a fetch, set up when its stream is identified
+         *
+         * @details Only fetch handlers ever have this. It belongs to the handler rather than the
+         *      stream carrying the fetch because the group order it decodes against is the one
+         *      the handler was created with.
          */
-        virtual void TryParseStreamBufferData(StreamContext& stream);
+        std::optional<messages::FetchObjectSerializationState> fetch_state_;
```
RichLogan: This feels less ideal to have this here. Also `is_fetch_handler` (which has always been here) seems removable. I wonder if we can put this somewhere else - maybe in `RxParseState` by the subgroup state? Not saying change it just maybe something to think about.

src/session.cpp
```diff
@@ -1528,272 +1528,431 @@ namespace quicr {
...
+        if (static_cast<ControlMessageType>(*stream_type) == ControlMessageType::kSetup) {
+            rx_ctrl_stream_id_ = stream_id;
+            return true;
         }
 
         /*
-         * RX data queue may have more messages at time of this callback. Attempt to
-         *      process all of them, up to a max. Setting a max prevents blocking
-         *      of other streams, etc.
+         * A header type nothing can be made of is the peer's problem with this one stream, not with
```
RichLogan: I'm not sure this is correct, this is a `PROTOCOL_VIOLATION` and should be treated as such.

src/session.cpp
```diff
@@ -1528,272 +1528,431 @@ namespace quicr {
...
+    }
+
+    bool Session::BindRecvStream(Stream& stream)
+    {
+        const auto stream_id = stream.GetStreamId();
+
+        // Peeked rather than consumed, so the header is still there for whoever the stream belongs
+        // to. Both kinds of data stream name their destination right after the type.
+        std::optional<std::uint64_t> stream_type;
+        std::optional<std::uint64_t> destination;
```
RichLogan: May for clarity it might be worth calling out this is `track_alias` or `request_id`? Idk, I had to go look it up at least in order to figure out what "destination" was, although it makes sense.

src/session.cpp
```diff
@@ -1528,272 +1528,431 @@ namespace quicr {
...
+        bool processed = false;
+        try {
+            processed = is_request_stream ? ProcessRequestMessage(stream, msg_type, payload)
+                                          : ProcessCtrlMessage(msg_type, payload);
+        } catch (const std::exception& e) {
+            QUICR_LOGGER_ERROR(logger_,
+                               "Caught exception trying to process control message. (type={}, error={})",
+                               static_cast<int>(msg_type),
+                               e.what());
+            throw ProtocolViolationException(e.what());
```
RichLogan: Without actually looking too hard, this feels like it might be overly loose, it could misclassify an internal error as a protocol violation and cause connection close. That being said, an exception in control handling is probably unrecoverable regardless, but semantically there might be a difference? In theory I guess that we could fail a given request, in an internal local error case, but keep the connection up for others. Again, just thinking on reading this.

src/transport_picoquic.cpp
```diff
@@ -1996,71 +1981,46 @@ try {
...
+         * The delegate stops short of draining a stream so a burst on one does not starve the
+         * others, and says when it has, earning another turn at the back of the queue. Claiming
+         * that turn before giving up this one leaves the sweep no gap to reclaim the buffer in
+         * while there is still something to read out of it.
+         */
+        if (connection->OnRecvStream(stream)) {
+            NotifyStreamRecv(connection, stream);
         }
+
+        stream->rx_notify_delivering.store(false);
```
RichLogan: Possibly this be in a defer and moved above the delegate call?

src/transport_picoquic.cpp
```diff
@@ -2087,17 +2047,17 @@ PicoQuicTransport::NotifyDgramRecv(const std::shared_ptr<PicoQuicConnection>& co
 
 void
 PicoQuicTransport::OnStreamClosed(const std::shared_ptr<PicoQuicConnection>& connection,
-                                  uint64_t stream_id,
-                                  std::shared_ptr<StreamRxContext> rx_ctx,
+                                  const std::shared_ptr<PicoQuicStream>& stream,
```
RichLogan: Since both of these are always retained for the notifier, perhaps they could be value type shared_ptrs in the signature and moved internally.

include/quicr/handlers/forwarding_subscribe_track_handler.h
```diff
@@ -0,0 +1,79 @@
...
+     *      wanting either of those wants a plain `SubscribeTrackHandler`.
+     *
+     *      Only streams are passed on as bytes, a datagram being a whole object with no stream to
+     *      pass on. Datagrams are still reported to `ObjectReceived` and still counted, and which
+     *      of the two an object arrives as is the publisher's to choose per object, so a track can
+     *      use both at once. Overriding `ObjectReceived` on one of these therefore reports the
+     *      track's datagrams and only its datagrams, and leaving it alone drops them; a relay
+     *      whose publisher may send any wants it implemented.
+     */
+    class ForwardingSubscribeTrackHandler : public SubscribeTrackHandler
```
RichLogan: I think we could probably make the docs for this more succinct.

src/transport_picoquic.cpp
```diff
@@ -1977,51 +1971,64 @@ try {
         return;
     }
 
-    rx_ctx->data_queue.Push(std::make_shared<const std::vector<uint8_t>>(bytes.begin(), bytes.end()));
+    {
+        std::lock_guard _(stream->rx_mutex);
+        stream->rx_data.Push(bytes);
```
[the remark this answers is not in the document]
RichLogan: This is still true right? Although we should probably do a separate DOS/resource limit pass everywhere.

src/transport_picoquic.cpp
GhostofCookie: Since it is going on the notify queue, I think maybe instead these might need to be copied and moved into the lambda instead.

PR #941, opened 2026-09-15
Move fetch eval first

RichLogan:
Fetch has to be checked before subscribe, because fetch handlers are subscribe handlers. The test didn't catch it because it didn't re-use the bad path.


TimEvens (approved):
LGTM

PR #942, opened 2026-09-17
Flush TX on subgroup abandon

RichLogan:
We're currently not flushing the TX queue on EndSubgroup(completed=false). I'm thinking we should be - otherwise the RESET to abandon this data will be queued behind the data it is signalling to abandon?


GhostofCookie (approved):

RichLogan:
Instead of this we might want to just close stream instead of enqueue the reset. That solves the tx object expiry issue (in the reset case at least).

RichLogan:
See #947 for probably the better fix

RichLogan:
See #947

PR #944, opened 2026-09-18
Performance updates

TimEvens:
* Updates picoquic to latest master with updated changes for XDP, recvmmsg and reduced poll calls
* Change to use queue instead of safequeue
* Reduce mark active


RichLogan (commented):
This looks good, although I opened https://github.com/Quicr/libquicr/pull/948 into this with a small tweak to the tx_mutex usage to protect tx_object and make the contract more explicit, although that's true even in the original code. 

The empty/push comment is true I think. Could we just wake unconditionally on push? If it's expensive I guess we need some more thinking there.

src/picoquic_connection.cpp
```diff
@@ -56,7 +57,7 @@ quicr::PicoQuicConnection::ReportMetricsSample(const MetricsTimeStamp& sample_ti
 
 quicr::PicoQuicStream::PicoQuicStream(const std::uint64_t stream_id,
                                       const std::uint64_t conn_id,
-                                      std::unique_ptr<SafeTimeQueue<ConnData>> tx_queue)
+                                      std::unique_ptr<std::queue<ConnData>> tx_queue)
```
GhostofCookie, line 60: Do we want to push this? We lose expiry if we do. Have we verified with using regular or even fast time_queue without the locking, given the stream locking should keep this safe now?

GhostofCookie (commented):
I'd hold off on merging this to main with std::queue replacing TQ, as we'd lose expiry.

RichLogan (dismissed):

PR #946, opened 2026-09-18
Serialize reset/close operation with enqueue 

RichLogan:
I believe there's a race today in terms of marking a stream RESET and potentially then enqueuing data onto it, which can cause an exception that bubbles up to a connection close, when it should be more gracefully handled.

`tx_closed` needs to be mutated under the `tx_data` lock for set, and also on the read in enqueue, in order to guarantee the proper serialised order of reset across a concurrent close vs. enqueue call, otherwise enqueue can see that `tx_closed` is false when it starts running, but it can subsequently become `true` during execution, and then we throw because the stream has gone. Taking the lock before checking ensures that doesn't happen and any failure will be a more graceful return rather than a later exception when it comes to see the enqueued data is for a stream that doesn't exist. 

All we really need to do is:

- Move the lock higher in enqueue
- External callers always lock tx_data before setting the atomic

`tx_data` could become private here and have an accessor too, to stop any misuse of the atomic set in the wrong way, but there's no misuse today. 

There is another side to this, which is that we probably should not attempt to enqueue for the closed subgroup streams in the first place, but that is a different change and this protection is still valid.


src/picoquic_connection.h
```diff
@@ -70,6 +70,18 @@ namespace quicr {
...
+        /**
+         * Close the send direction.
+         */
+        void MarkTxClosed()
+        {
+            if (tx_data == nullptr) {
+                return;
+            }
+            std::lock_guard _(*tx_data);
+            tx_closed.store(true, std::memory_order_release);
```
GhostofCookie, line 82: What good does locking the queue do here exactly? Is this just so that it blocks writers from pushing more onto it? Should we purge it here?

src/transport_picoquic.cpp
```diff
@@ -1216,10 +1216,16 @@ PicoQuicTransport::EnqueueStream(const std::shared_ptr<PicoQuicConnection>& conn
                                  const uint32_t ttl_ms,
                                  const EnqueueFlags flags)
 {
+    if (stream == nullptr || stream->tx_data == nullptr) {
+        return TransportError::kInvalidStreamId;
+    }
+
+    std::lock_guard _(*stream->tx_data);
```
GhostofCookie, line 1223: I wonder, is this a place where we should be locking the whole stream?

src/picoquic_connection.h
RichLogan: It's specifically to serialise altering the tx_closed value here against any concurrent enqueue operation that relies on this value being set. If we're here, writing tx_closed = true, we want to make sure that Enqueue cannot be running concurrently to avoid the case where it read tx_closed = false, this tx_closed = true, and then rest of enqueue runs. Thus we take this mutex, not because we need to protect tx_data specifically, but because we need some lock between enqueue and this to synchronise the close.

src/transport_picoquic.cpp
RichLogan: I think this is probably true, but that might be a slightly bigger change because we probably then want to not have tx_data be lockable since we'd be protecting it via the stream lock.

TimEvens (approved):
LGTM

PR #947, opened 2026-09-18
Immediate reset on incomplete subgroup

RichLogan:
This probably the proper fix over #942 

I didn't really want to expose a generic `CloseStream` to `Session` if I could avoid it which is why I specifically exposed it as `ResetSubgroup`. 

A test for this would be nice but without mocking a fake transport that doesn't egress bytes it's a bit awkward.


TimEvens (approved):
LGTM

